PR #17, opened 2017-11-10
Add back logRawData fn

samuelli:


src/metrics/review-latency.ts
```diff
@@ -35,6 +35,39 @@ export class ReviewLatencyResult {
     return `There were ${this.reviews.length} reviews ` +
         `with an average latency of ${avg} hours.`;
   }
+
+
+  /**
+   * Logs raw data for review latency. Currently logs average review latency by
+   * weeks starting on Sunday.
+   */
+  logRawData() {
```
aomarks: Wondering if metrics should typically have a function that returns data objects (e.g. here `timeSeries(): {date: string, latency:number}`) and then we can do the CSV transformation separately. This is good for now though IMO.

PR #19, opened 2017-11-11
Add issue counts metric.

aomarks:
Also some refactoring.


src/common-queries.ts
```diff
@@ -0,0 +1,70 @@
...
+ * the License.
+ */
+
+import gql from 'graphql-tag';
+
+import {GitHub} from './gql';
+import * as gqlTypes from './gql-types';
+
+/**
+ * Fetch the repos under the given GitHub org/person.
```
samuelli: Only for orgs, a different endpoint has to be used if its a person.

src/gql.ts
```diff
@@ -1,15 +1,20 @@
...
 import fetch from 'node-fetch';
 import {promisify} from 'util';
 
+const schema = require('../github-schema.json');
+
 export class GitHub {
   private apollo: ApolloClient<NormalizedCache>;
 
   constructor(uri = 'https://api.github.com/graphql') {
+    const fragmentMatcher = new IntrospectionFragmentMatcher(
```
samuelli: Maybe a quick comment on why this is needed since it's a little odd looking?

src/metrics/issue-counts.ts
```diff
@@ -0,0 +1,206 @@
...
+            closedAt = ev.createdAt;
+            break;
+          }
+        }
+        if (!closedAt) {
+          // Edge case where we couldn't find the closed event even though
+          // we're closed. This could happen if there was a lot of continuing
+          // discussion after an issue was closed. Skip these issues
+          // altogether.
+          console.log('Missing closed date for closed issue:', issue.url);
```
samuelli: Let's remove this console.log(). I was imagining it would be useful during testing to ensure we had the right thresholds, but after that its fine to just skip them.

src/common-queries.ts
aomarks: Done.

src/metrics/issue-counts.ts
aomarks: Done.

src/gql.ts
aomarks: Done.

PR #20, opened 2017-11-13
Improvements to replay server.

aomarks:
- Make the replay file format easier to read by writing indented JSON. Also gets rid of the wrapper object so there is no double-escaping.

- Make the query variables hash less brittle. Previously it was just a JSON encoding, which doesn't have a deterministic property ordering, and also distinguished between null/undefined which we don't want.

- Forward the HTTP headers sent by the Apollo client instead of writing our own, and use gzip.


samuelli:
Awesome!

PR #23, opened 2017-11-15
Add a review coverage metric

samuelli:


README.md
```diff
@@ -42,8 +42,6 @@ A few tips when creating a new metric:
 ## Testing
 To run the tests, use `npm test`. The tests use a mock server which emulates
 the GitHub API by allowing you to record GitHub API responses. To add a new
-test, write your test then run the following: ``` npm run test:record ```
-
-To avoid re-recording repsonses for other tests, use Ava's `test.only` when
-recording. Check the generated files so that tests will use these snapshotted
-responses.
+test, write your test then run the following: ``` npm run test:record --
+--match 'testname*``` Check in the generated files so that tests will use
```
aomarks: 'testname*'

src/metrics/review-coverage.ts
```diff
@@ -0,0 +1,293 @@
...
+import {getOrgRepos} from '../common-queries';
+import {GitHub} from '../gql';
+import {PullRequestCommitsQuery, PullRequestCommitsQueryVariables} from '../gql-types';
+import {RepoCommitsQuery, RepoCommitsQueryVariables} from '../gql-types';
+import {RepoPRsCommitsQuery, RepoPRsCommitsQueryVariables} from '../gql-types';
+
+import {getReviewsForPullRequest} from './review-latency';
+
+export class ReviewCoverageResult {
+  commits: Map<string, ReviewedCommit>;
```
aomarks: Comment what key is? Or just make it an array? Doesn't look like it needs to be a map here.

src/metrics/review-coverage.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  constructor(commits: Map<string, ReviewedCommit>) {
+    this.commits = commits;
+  }
+
+  summary() {
+    const count = this.numReviewed();
+    return `There are ${this.commits.size} commits of which ${
```
aomarks: I find this formatting kinda strange, it sort of cheats by breaking in the middle of expressions. Think I prefer manually breaking into 3 strings.

src/metrics/review-coverage.ts
```diff
@@ -0,0 +1,293 @@
...
+        count} are reviewed. \nReview coverage is ${
+        Math.round(count / this.commits.size * 100)}%.`;
+  }
+
+  numReviewed(): number {
+    let numReviewed = 0;
+    this.commits.forEach((x) => x.reviewed ? numReviewed++ : '');
```
aomarks: Maybe prefer a regular for loop here? I think the inert empty string negative case of the ternary is a little funny to read. Could also do `numReviewed += x.reviewed ? 1 : 0` to use the ternary more straightforwardly? Also you could do `forEach(({reviewed}) => ... }`

src/metrics/review-coverage.ts
```diff
@@ -0,0 +1,293 @@
...
+    return numReviewed;
+  }
+}
+
+type ReviewCoverageOpts = {
+  org: string,
+  repo?: string,
+  since?: string,  // ISO date string
```
aomarks: Should `since` be non-optional?

src/metrics/review-coverage.ts
```diff
@@ -0,0 +1,293 @@
...
+};
+
+type Commit = {
+  oid: string,
```
aomarks: What's oid?

src/metrics/review-coverage.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  const commits: Map<string, ReviewedCommit> = new Map();
+  for (const {owner, name} of repos) {
+    // Get all commits on the master branch.
+    for (const commit of await getMasterCommits(github, owner, name)) {
+      commits.set(commit.oid, Object.assign({reviewed: false}, commit));
+    }
+
+    // Mark all reviewed PR commits as reviewed.
+    for (const commit of await getPRCommitsForRepo(github, owner, name)) {
```
aomarks: Maybe swap the order of these for loops, so you first make a `Set<string>` with all the reviewed oids? Then you could have:

`Object.assign({reviewed: reviewedCommits.has(oid)}, ...)`,  so you wouldn't need the `!`, and you don't need to set the reviewed state twice.

src/metrics/review-coverage.ts
```diff
@@ -0,0 +1,293 @@
...
+ * Get all the commits on the default (normally master) branch for the last
+ * year.
+ */
+async function getMasterCommits(
+    github: GitHub, owner: string, name: string, since?: string):
+    Promise<Commit[]> {
+  const results =
+      github.cursorQuery<RepoCommitsQuery, RepoCommitsQueryVariables>(
+          repoCommitsQuery, {owner, name, since}, (data) => {
+            if (!data.repository || !data.repository.defaultBranchRef ||
```
aomarks: Maybe pull this into a `getPageInfo` since it's kinda long to be inline.

src/metrics/review-latency.ts
```diff
@@ -148,7 +148,7 @@ type Review = {
  * must not be from the pull request author and must be from different
  * reviewers.
  */
-function getReviewsForPullRequest(pullRequest: PullRequest): Review[] {
+export function getReviewsForPullRequest(pullRequest: PullRequest): Review[] {
```
aomarks: Factor out to a shared file? Maybe `common-queries` should just be `common` for now.

README.md
samuelli: Hawk eye. Nice; fixed.

src/metrics/review-coverage.ts
samuelli: Done, changed to array.

src/metrics/review-coverage.ts
samuelli: Done.

src/metrics/review-coverage.ts
samuelli: Changed to the ternary with `? 1 : 0`. Destructuring is fancy, changed, though I'm not sure about its readability since now there's just lots of parens.

src/metrics/review-coverage.ts
samuelli: I left is as optional since its a pain to pass in, and you may want for all time anyway. This param is only helpful if your repo is older than when reviews started, and you want better results there.

src/metrics/review-coverage.ts
samuelli: Some object id that GitHub uses for their git object ids. It differs from GitHub's normal IDs to identify their objects.

src/metrics/review-coverage.ts
samuelli: Done.

src/metrics/review-coverage.ts
samuelli: I thought you said this doesn't get typechecked properly if its not inline?

samuelli: Done.

src/metrics/review-latency.ts
samuelli: Done, done.

src/metrics/review-coverage.ts
aomarks: Yeah, tbh I kinda prefer a regular for loop and if here

samuelli: Done

samuelli:
@aomarks: PTAL

PR #24, opened 2017-11-17
Add test coverage

samuelli:
This change adds tests coverage reporting. It uses [nyc](https://github.com/istanbuljs/nyc) which is based on [instanbul](https://github.com/gotwarlost/istanbul) with the additional support for spawned subprocesses, which is what ava does during testing.

[Coveralls](https://coveralls.io/) allows for coverage reporting after Travis runs coverage and shows up as a badge on the repo. [Detailed reports](https://coveralls.io/github/PolymerLabs/project-health) are available on Coveralls site.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,5 @@ dist: trusty
 cache:
   directories:
     - node_modules
+after_script:
+  - npm run report-coverage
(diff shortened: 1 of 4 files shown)


aomarks:
Cool. So it will actually fail test if coverage drops below the amount you configured here? Is that on the incremental diff, or overall?

samuelli:
> Cool. So it will actually fail test if coverage drops below the amount you configured here? Is that on the incremental diff, or overall?

Yes, as its part of the test command, it'll fail on Travis if it drops below the defined amounts there. It's for the overall code base. Coveralls also supports a status check fail below certain levels that can be configured from their site.

PR #28, opened 2017-11-29
Add stars metric

samuelli:
This adds a stars metric which gets the raw star data to allow for star trends to be mapped (not currently exposed yet).

### DIFF
--- a/src/metrics/stars.ts
+++ b/src/metrics/stars.ts
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import gql from 'graphql-tag';
+
+import {getOrgRepos} from '../common';
+import {GitHub} from '../gql';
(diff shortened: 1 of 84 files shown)


samuelli:
This could be implemented in a different way to fetch total stars much more efficiently, but it would not allow us to analyze stars over time, which would be useful to correlate metrics.

src/metrics/stars.ts
```diff
@@ -0,0 +1,111 @@
...
+  constructor(stars: Star[]) {
+    this.stars = stars;
+  }
+
+  summary(): string {
+    return `There are ${this.stars.length} stars.`;
+  }
+}
+
+type StarOpts = {
```
aomarks: `StarsOpts` I think

src/metrics/stars.ts
```diff
@@ -0,0 +1,111 @@
...
+  const results =
+      github.cursorQuery<StarsQuery, StarsQueryVariables>(
+          starsQuery,
+          {owner, name},
+          (data) => data.repository && data.repository.stargazers);
+
+  for await (const data of results) {
+    const repo = data.repository;
+    if (!repo) {
+      break;
```
aomarks: `break` or `continue`?

src/test/metrics/stars-test.ts
```diff
@@ -0,0 +1,26 @@
...
+  const {server, client} = await startTestReplayServer(t);
+  t.context.server = server;
+  t.context.client = client;
+});
+
+test.afterEach.cb((t) => {
+  t.context.server.close(t.end);
+});
+
+test('WebComponents stars', async (t) => {
```
aomarks: I think you need to `git add` the test replay files.

aomarks: Oh sorry, just saw your comment. SG.

samuelli: Done

src/metrics/stars.ts
samuelli: Done.

src/metrics/stars.ts
samuelli: `continue`. Searching, there's a few other instances where this looks like its probably wrong. Going to send a separate PR to fix.

PR #34, opened 2017-12-04
Add test data for UI testing

samuelli:


src/server/server.ts
```diff
@@ -104,6 +104,53 @@ async function fetchUserData(token: string) {
 
 app.use(cookieParser());
 
+/**
+ * Merges two objects that have the same structure. It always folds into the
+ * first argument and always concats any array that is found in the tree.
+ */
+export function mergeObjects(
```
aomarks: Maybe the signature can be something like `mergeObjects<T extends {[key: string]: any}>(left: T, right: T): T`?

samuelli: Done.

PR #38, opened 2017-12-14
Refactor client/server structure

samuelli:
This change refactors the existing project into 2 environments: client & server. Most files moved into the server folder with the exception of static files which have been moved to client. This will allow client to contain its own TS environment and manage its own dependencies. Paths, readme & .travis.yml have been updated to match the change.


aomarks:
The `server/package.json` still depends on `lit-html`. I think I expected to also see a new `client/package.json` with the client deps.

samuelli:
@aomarks: plan on adding client/package.json in a separate change.

PR #39, opened 2017-12-14
Add tests for dashboard server

samuelli:
This change uses dependency injection to pass the GitHub client into the dashboard server to allow for testing. The test then re-uses the recording/replay server for testing against the GitHub API. The test uses a test account `project-health1` so that tests are real, but unchanging. To re-record the API responses, any GitHub token must be used with the test user still fixed.


server/src/server.ts
```diff
@@ -12,8 +11,123 @@ import {DashResponse, PullRequest} from '../../api';
 import {GitHub} from './gql';
 import {ViewerLoginQuery, ViewerPullRequestsQuery} from './gql-types';
 
-const app = express();
-const github = new GitHub();
+export class DashServer {
+  github: GitHub;
```
aomarks: Can these be private?

samuelli: Done

PR #47, opened 2017-12-28
Add util function to figure out top repos by user

samuelli:


server/src/github.ts
```diff
@@ -20,8 +21,11 @@ const schema = require('../github-schema.json');
 
 export class GitHub {
   private apollo: ApolloClient<NormalizedCacheObject>;
+  private jsonUrl: string;
 
-  constructor(uri = 'https://api.github.com/graphql') {
+  constructor(
+      uri = 'https://api.github.com/graphql',
```
aomarks: Maybe call this `gqlUrl` to distinguish it from the new `jsonUrl`.

server/src/github.ts
```diff
@@ -108,6 +114,28 @@ export class GitHub {
       yield result.data;
     }
   }
+
+  // tslint:disable-next-line:no-any
+  async get(path: string, userToken: string, parseJSON = true): Promise<any> {
```
aomarks: Why would we ever want `parseJSON = false`?

aomarks: I think you could return `Promise<object|string>`.

server/src/test/replay-server.ts
```diff
@@ -78,26 +101,46 @@ export async function startTestReplayServer(t: ava.TestContext):
...
-        res.end(replayBody);
+        await sendReplayFile(res, replayFile);
       }
     });
   }
 
+  async function jsonHandler(
+      req: http.IncomingMessage, res: http.ServerResponse) {
+    const replayFile =
+        path.join(replayDir, (req.url || '').replace(/\//g, '#'));
```
aomarks: Took me a second -- add a comment like "replace path delimiters with arbitrary hash character so that we create a flat file structure".

server/src/test/util/my-repos-test.ts
```diff
@@ -0,0 +1,29 @@
+import test from 'ava';
+
+import {getMyRepos} from '../../util/my-repos';
+import {startTestReplayServer} from '../replay-server';
+
+test.beforeEach(async (t) => {
+  const {server, client} = await startTestReplayServer(t);
+  t.context.server = server;
```
aomarks: Use the type safe ava context setup from that other recent PR?

server/src/util/my-repos.ts
```diff
@@ -0,0 +1,150 @@
...
+ */
+
+import gql from 'graphql-tag';
+import {GitHub} from '../github';
+import {MyReposQuery, MyReposQueryVariables} from '../gql-types';
+
+const MAX_STATS_RETRIES = 5;
+
+// Variables that impact the scoring algorithm.
+const SCORE_THRESHOLD = 20;
```
aomarks: Maybe an extra sentence for SCORE_THRESHOLD and CONTRIBUTION_SIZE?

server/src/util/my-repos.ts
```diff
@@ -0,0 +1,150 @@
...
+ * based on contribution recency and consistency.
+ */
+export async function getMyRepos(
+    github: GitHub, login: string, userToken: string): Promise<string[]> {
+  const repos = new Map<string, number>();
+  const results = github.cursorQuery<MyReposQuery, MyReposQueryVariables>(
+      reposQuery, {login}, (q) => q.user && q.user.contributedRepositories);
+  const promises = [];
+  for await (const data of results) {
+    if (data.user) {
```
aomarks: Would prefer `if (!data.user) { continue; }` and same for `repo` below, so that the normal flow has minimal indentation.

server/src/util/my-repos.ts
```diff
@@ -0,0 +1,150 @@
...
+  // Filter out repos with small levels of contributions.
+  result = result.filter((key) => (repos.get(key) || 0) > SCORE_THRESHOLD);
+  return result.sort(comparator);
+}
+
+/**
+ * Given a repo and a user, returns a weighted contribution score for the repo
+ * (0-100).
+ */
+function getContributionWeight(
```
aomarks: Make this an `async` function? Looks like it should work fine that way, but maybe I missed something. To reject, just `throw`.

server/src/util/my-repos.ts
```diff
@@ -0,0 +1,150 @@
...
+// Based on API response documented at
+// https://developer.github.com/v3/repos/statistics/#get-contributors-list-with-additions-deletions-and-commit-counts
+type GitHubStatsResponse = [{
+  author: {login: string},
+  total: number,
+  weeks: [{w: string, a: number, d: number, c: number}],
+}];
+
+async function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
```
aomarks: `const sleep = util.promisify(setTimeout)` (I would maybe call it `setTimeoutPromise` but up to you).

server/src/util/my-repos.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+`;
+
+// Based on API response documented at
+// https://developer.github.com/v3/repos/statistics/#get-contributors-list-with-additions-deletions-and-commit-counts
+type GitHubStatsResponse = [{
+  author: {login: string},
+  total: number,
+  weeks: [{w: string, a: number, d: number, c: number}],
```
aomarks: It's at the link, but would be nice to add a comment here too: (w)eek unix timestamp, (a)dditions, (d)eletions, (c)ommits

server/src/util/my-repos.ts
```diff
@@ -0,0 +1,150 @@
...
+        if (stat.length === 0) {
+          resolve(0);
+          return;
+        }
+
+        let contributions = 0;
+        for (const week of stat[0].weeks) {
+          // Convert UNIX timestamp in seconds to Date object.
+          const date = new Date(Number(week.w) * 1000);
+          const timeSince = Math.max(
```
aomarks: Comment (or rename the var) what's the unit for `timeSince`?

aomarks: Comment (or rename the var) what's the unit for `timeSince`?

server/src/util/my-repos.ts
```diff
@@ -0,0 +1,150 @@
...
+                                (new Date().getTime() - date.getTime()),
+                                CONTRIBUTION_WINDOW) /
+              CONTRIBUTION_WINDOW;
+          // Produce a weight between 0-1 based on the weeks passed. Anything
+          // less than CONTRIBUTION_WINDOW will be ~1.
+          const timeWeight = Math.tanh(5 / timeSince);
```
aomarks: Comment why 5?

server/src/util/my-repos.ts
```diff
@@ -0,0 +1,150 @@
...
+          // Weight any contribution by recency. Delta sizes are ignored.
+          if (week.a + week.d + week.c) {
```
aomarks: The `+` threw me a bit since the result isn't used, maybe use `||` instead?

server/src/github.ts
samuelli: Done.

server/src/github.ts
samuelli: I use it in this PR, where reading the status code response is important to determining the type of response. The `parseJSON` flag is passed along to `request-promise-native`, which auto parses successful responses and throws a rejection on any non-200 status code.

samuelli: It's not a string, its a response object from the request library, so I'm not sure how to type it. `request.get()` returns an any.

server/src/test/replay-server.ts
samuelli: Done.

server/src/test/util/my-repos-test.ts
samuelli: Done.

server/src/util/my-repos.ts
samuelli: Done.

server/src/util/my-repos.ts
samuelli: Done.

server/src/util/my-repos.ts
samuelli: Oh, sg. Done.

server/src/util/my-repos.ts
samuelli: Done.

server/src/util/my-repos.ts
samuelli: Not sure how to better comment this. It's based on the comment. If it was 1, it would produce ~0.76 not 1 for the CONTRIBUTION_WINDOW.

server/src/util/my-repos.ts
samuelli: Done.

server/src/util/my-repos.ts
samuelli: Done.

server/src/util/my-repos.ts
samuelli: Done, done.

PR #48, opened 2018-01-02
Adding push subscribing to a push template

gauntface:
- Made some minor changes to the server to get it to run locally, not 100% sure if this was the correct fix though.
- Adds a new template for enabling / disabling notifications. At the moment this is just on the home page. Chances are we'll want to move this to somewhere else depending on our UI choices, but probably ok for now.
    - If you want to add this to the dash template I can do so, I didthis since it seems a logical split of concerns.
- I may be missing some tricks with lit-html.


client/push.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (isUnsubscribed) {
+    console.log('TODO: Must remove subscription from backend', subscription);
+  } else {
+    console.log('TODO: Add subscription to backend', subscription);
+  }
+}
+
+/**
+ * The pushManage.subscribe() method expects an ArrayBuffer and not a string.
+ * This method convertys a base64 string into a Uint8Array.
```
samuelli: converts

client/push.ts
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * (Note: The spec has changed to allow strings, but browser suppoer is lacking)
```
samuelli: support

client/push.ts
```diff
@@ -0,0 +1,122 @@
...
+ * @param base64String
```
samuelli: I've been removing these since we don't need the jsdoc for anything.

client/push.ts
```diff
@@ -0,0 +1,122 @@
...
+  updateUI();
+
+  await updateBackend(isUnsubscribed, subscription);
+
+  button.removeAttribute('disabled');
+}
+
+/**
+ * This method is called when the UI is updated. It ensures the correct
+ * message is displayed to the user to enable or disable notifications,
```
samuelli: s/,/./

PR #49, opened 2018-01-03
Adding an oauth page

gauntface:
- Adds an oauth page which opens up any changes to the sign in flow we want to add in the future.
- Moves building and deploying of the whole project to a top level package.json file
- Enables use of localhost or remote url as the final dash.html page after sign in.


server.js
```diff
@@ -3,5 +3,14 @@
...
 const {GitHub} = require('./server/lib/github');
 
+try {
+  const secrets = require('./secrets.json');
+  if (secrets['GITHUB_CLIENT_SECRET']) {
+    process.env.GITHUB_CLIENT_SECRET = secrets['GITHUB_CLIENT_SECRET'];
+  }
+} catch (err) {
+  // No secrets defined.
+}
```
samuelli: Probably should remove the try catch or throw the error? The app can't really run without these secrets, so best to cause an exception.

client/oauth.html
```diff
@@ -0,0 +1,25 @@
+<script>
+  const CLIENT_ID = `23b7d82aec29a3a1a2a8`;
```
samuelli: Change to single quotes. Avoid backticks generally if there's no expression.

server/src/dash-server.ts
```diff
@@ -12,11 +12,19 @@ import {GitHub} from './github';
 import {ViewerLoginQuery, ViewerPullRequestsQuery} from './gql-types';
 
 export class DashServer {
+  private secrets: {
+    GITHUB_CLIENT_ID: string,
+    GITHUB_CLIENT_SECRET: string,
+  };
```
samuelli: Fine for now, but you can add a type here so you don't have to type it out everytime.

PR #52, opened 2018-01-05
Push model

gauntface:
Adds endpoint `/api/push-subscriptions/add` to accept push subscriptions and the corresponding supported content encoding from the client to the server.

Saves this to a model that will stash this in memory for now.

### DIFF
--- a/client/push.ts
+++ b/client/push.ts
@@ -1,9 +1,5 @@
 import {html, render} from './node_modules/lit-html/lit-html.js';
 
-declare var PushManager: {
-  supportedContentEncodings?: string[];
-};
-
 const PUBLIC_VAPID_KEY = 'BOX5Lqb44uosZL4_UtV7XW9dHaBj9ERFbCzlsYZBObMZjIB-yxPIbjI5pTBgIt09iy-Hl57AWpr7lJ6QXaQjy30';
 
 async function updateBackend(isUnsubscribed: boolean, subscription: PushSubscription) {
@@ -12,7 +8,8 @@ async function updateBackend(isUnsubscribed: boolean, subscription: PushSubscrip
   } else {
     const bodyContent = {
       subscription: subscription.toJSON(),
-      supportedContentEncodings: PushManager.supportedContentEncodings || [],
+      supportedContentEncodings:
+        (PushManager as {supportedContentEncodings?: string[]}).supportedContentEncodings || [],
     };
 
     await fetch(`/api/push-subscription/add`, {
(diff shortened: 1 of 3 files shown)


client/push.ts
```diff
@@ -1,12 +1,28 @@
 import {html, render} from './node_modules/lit-html/lit-html.js';
 
+declare var PushManager: {
```
aomarks: This is actually not augmenting the global `PushManager` variable. Since this is an ES module, the `var` is not globally scoped, it's scoped to the module, so you're declaring that there will be a different variable called `PushManager`. I guess it will actually work, because no JS will be emitted for this `declare` statement, so when you reference it below you'll end up referencing the global symbol anyway, but it feels kind of wrong to trick the compiler this way.

To add a declaration to the global scope from a module, you have to do it explicitly like this:

```
declare global {
  var foo ...
}
```

However, this won't work for us here because you actually just can't do declaration merging with variables (if you try it, it will say you can't re-declare `PushManager` with a different type).

IMO the TypeScript declarations seem wrong. `var PushManager` is actually an instance with type `interface PushManager`, but they way they are written (https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L9142) suggests that `var PushManager` is a constructor that returns a new `interface PushManager`. Maybe we should submit a PR to fix it (as well as adding `supportedContentEncodings`).

Sadly, for now I think you should just do `(PushManager as any).supportedContentEncodings` and add a comment.

server/src/models/PushSubscriptionModel.ts
```diff
@@ -0,0 +1,33 @@
...
+    auth: string;
+  };
+}
+
+class PubscriptionInfo {
+  subscription: PushSubscription;
+  supportedContentEncodings: string[];
+}
+
+class PushSubscriptionModel {
```
samuelli: I think you can just have `export class PushSubscriptionModel`.

server/src/dash-server.ts
```diff
@@ -84,6 +89,31 @@ export class DashServer {
...
   }
 
+  async addPushSubscription(req: express.Request, res: express.Response) {
+    if (!req.body) {
+      res.sendStatus(400);
+      return;
+    }
+
+    const token = req.cookies['id'];
+    const loginResult = await this.github.query<ViewerLoginQuery>(
```
samuelli: We should generally avoid this at some point. When we first exchange for a token, we should store the user login with the token somewhere. Maybe just add a TODO here about not having to do this query again?

PR #57, opened 2018-01-09
Support sending token to localhost

gauntface:
I noticed during local development that the github code wasn't actually getting sent to localhost, instead my machine was using the environment github token.

This should fix that.


client/oauth.html
```diff
@@ -2,24 +2,31 @@
...
 
-  var queryParams = new URLSearchParams(window.location.search);
+  (function() {
+    var queryParams = new URLSearchParams(window.location.search);
+    const redirectOrigin = queryParams.get('redirect-origin');
+    // This redirect allows testing on localhost or on the remote server
+    if (redirectOrigin) {
+      // Redirect to the same URL, just on the different origin.
+      const redirectOauthURL = new URL(window.location.pathname, redirectOrigin);
+      redirectOauthURL.search = `code=${loginCode}`;
```
samuelli: loginCode isn't defined yet.

gauntface:
@samuelli should be patched now.

PR #59, opened 2018-01-16
Starting to implementing dashboard design

gauntface:
Starting to hook up some of the dashboard designs.

dashboard-screenshot

In the next PR I'll move out the push notification button out into a settings page.

Changes:
- Adding values to `PullRequest` type to fill out the UI with correct info.
- Moved some styles that weren't specific to the component out into the html file.
- Altered the dash template to use a css grid layout.
- Moved to CSS variables to encourage sharing of some of the styles within a component.
- Moved to a BEM-esque naming pattern.


client/dash.ts
```diff
@@ -7,59 +7,156 @@ async function start() {
...
-          </div>
-          <div>
-            Line 2
-          </div>
-        </div>
+  const pullRequestTemplate = (pr: PullRequest) => {
+    let approverElements = ``;
+    if (pr.approvedBy.length > 0) {
+      approverElements = `
+      <span class="pr-status__seperator">&middot;</span>
```
samuelli: separator*

client/dash.ts
```diff
@@ -7,59 +7,156 @@ async function start() {
...
-        <div class="footer">
-          <div>
-            Approved by nobody
-          </div>
-          <div>
-            Line 2
-          </div>
-        </div>
+  const pullRequestTemplate = (pr: PullRequest) => {
+    let approverElements = ``;
```
samuelli: Can we use '' here instead?

client/dash.ts
```diff
@@ -7,59 +7,156 @@ async function start() {
...
+    if (pr.approvedBy.length > 0) {
+      approverElements = `
+      <span class="pr-status__seperator">&middot;</span>
+      <span class="pr-status__approvals">approved by ${
+          pr.approvedBy.join(', ')}</span>
+      `;
+    }
+
+    let ellapsedValue = (Date.now() - pr.createdAt) / 1000;
+    let ellapsedUnit = 'seconds';
```
samuelli: elapsed*

api.d.ts
```diff
@@ -2,12 +2,16 @@ export interface PullRequest {
   repository: string;
   title: string;
   number: number;
+  prUrl: string,
+  login: string,
```
samuelli: Maybe `author` instead? I think that would be clearer.

PR #61, opened 2018-01-18
Adding status info

gauntface:
This starts to add in contextual status info and determining if the PR requires user interaction or not.

dash-with-status

Things to note:
- I opted for "Changes requests" instead of "Waiting on you".
- This only accounts for reviews, not status checks.
- The next step can be to add the event points on the grid.

### DIFF
--- a/api.d.ts
+++ b/api.d.ts
@@ -12,6 +12,7 @@ export interface PullRequest {
   pendingReviews: string[];
   statusState: 'passed'|'pending'|'failed';
   actionable: boolean;
+  actionMsg: string|null;
 }
 
 export interface DashResponse {
--- a/server/src/dash-server.ts
+++ b/server/src/dash-server.ts
@@ -9,7 +9,7 @@ import * as request from 'request-promise-native';
 import {DashResponse, PullRequest} from '../../api';
 
 import {GitHub} from './github';
-import {ViewerLoginQuery, ViewerPullRequestsQuery} from './gql-types';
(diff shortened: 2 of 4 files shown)


api.d.ts
```diff
@@ -12,6 +12,7 @@ export interface PullRequest {
   pendingReviews: string[];
```
samuelli: This data type was written when we didn't exactly have good mocks. I don't think this maps to what we need anymore. eg. we need to know when the review was approved. The UI now shows outgoing/incoming very differently and we should probably represent them differently.

server/src/dash-server.ts
```diff
@@ -155,6 +155,86 @@ export class DashServer {
...
+
+              if (prReview.state === PullRequestReviewState.APPROVED) {
+                approvedBy.push(prReview.author.login);
+              } else if (
+                  prReview.state === PullRequestReviewState.CHANGES_REQUESTED) {
+                changesRequestedBy.push(prReview.author.login);
+              } else if (prReview.state === PullRequestReviewState.COMMENTED) {
+                commentedBy.push(prReview.author.login);
+              } else if (prReview.state === PullRequestReviewState.PENDING) {
+                pendingReviews.push(prReview.author.login);
```
samuelli: I would probably keep a single array of objects, and filter at the end.

gauntface:
Gonna go a diff direction after chatting with @samuelli

PR #65, opened 2018-01-22
Demonstrate end-to-end firestore integration.

aomarks:
This adds a `/firestore-test` endpoint that shows all of the documents in the `test` collection in our firestore instance.

Running the server locally now requires a `github-health-staging-*.json` Google Cloud credentials file. When deployed to App Engine, this file is not needed because it will use Application Default Credentials.

Adds a new `pre-deploy.js` which:
- Re-compiles the grpc native extension to target the GAE architecture.
- Deletes any unnecessary files from `node_modules` to keep us away from the GAE file limit.


api.d.ts
```diff
@@ -14,9 +14,7 @@ export interface OutgoingPullRequest extends PullRequest {
   reviewRequests: string[];
 }
 
-export interface DashResponse {
-  outgoingPrs: OutgoingPullRequest[];
-}
+export interface DashResponse { outgoingPrs: OutgoingPullRequest[]; }
```
samuelli: Is this the auto formatter? Looks odd

package.json
```diff
@@ -5,18 +5,20 @@
   "private": true,
   "scripts": {
     "clean:client": "cd ./client/ && rm -rf ./node_modules && npm install --only=production",
-    "clean:server": "cd ./server/ && rm -rf ./node_modules && npm install --only=production && cd node_modules && echo num dep files before prune: $(find . -type f | wc -l) && find . -type f -not \\( -name '*.js' -or -name '*.json' -or -name '*.node' \\) -delete && echo num dep files after prune: $(find . -type f | wc -l)",
+    "clean:server": "cd ./server/ && rm -rf ./node_modules && npm install --only=production && cd - && npm run build-prod-grpc-extension && npm run prune-server-deps",
+    "build-prod-grpc-extension": "cd ./server/node_modules/grpc && node_modules/.bin/node-pre-gyp install --fallback-to-build --library=static_library --target_arch=x64 --target_platform=linux --runtime=node --target=8.9.0 --target_libc=glibc",
```
samuelli: This is kinda crazy. Maybe a comment would be helpful?

https://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install

server/src/dash-server.ts
```diff
@@ -28,6 +30,8 @@ export class DashServer {
     this.github = github;
     this.secrets = secrets;
     this.pushSubscriptions = new PushSubscriptionModel();
+    this.pushSubscriptions = new PushSubscriptionModel();
```
samuelli: This looks duplicated?

api.d.ts
aomarks: Yep, it's the auto formatter. It wasn't configured to format this file before this PR.

server/src/dash-server.ts
aomarks: Oops, fixed.

package.json
aomarks: Huh, didn't know that was valid. Added a comment for this and the `prune-server-deps` command.

samuelli: Thanks. It's a bit awkward, but in this case I think it needed a comment.

package.json
```diff
@@ -5,18 +5,32 @@
...
+      "to be deploying from the exact same architecture that we are deploying to, we need to ",
+      "recompile it for App Engine."
+    ],
+    "build-prod-grpc-extension": "cd ./server/node_modules/grpc && node_modules/.bin/node-pre-gyp install --fallback-to-build --library=static_library --target_arch=x64 --target_platform=linux --runtime=node --target=8.9.0 --target_libc=glibc",
+
+    "//": [
+      "We have a 20K file limit for App Engine. Our server node_modules gets us close to this ",
+      "limit. Delete all unnecessary files to buy us some time."
+    ],
+    "prune-server-deps": "cd ./server/node_modules && echo num dep files before prune: $(find . -type f | wc -l) && find . -type f -not \\( -name '*.js' -or -name '*.json' -or -name '*.node' -or -name '*.pem' -or -name '*.proto' \\) -delete && echo num dep files after prune: $(find . -type f | wc -l)",
```
gauntface: We no longer install files under `./server/node_modules/` only `./node_modules`.

package.json
```diff
@@ -5,18 +5,32 @@
...
     "clean:client": "cd ./client/ && rm -rf ./node_modules && npm install --only=production",
-    "clean:server": "cd ./server/ && rm -rf ./node_modules && npm install --only=production && cd node_modules && echo num dep files before prune: $(find . -type f | wc -l) && find . -type f -not \\( -name '*.js' -or -name '*.json' -or -name '*.node' \\) -delete && echo num dep files after prune: $(find . -type f | wc -l)",
+    "clean:server": "cd ./server/ && rm -rf ./node_modules && npm install --only=production && cd - && npm run build-prod-grpc-extension && npm run prune-server-deps",
+
+    "//": [
+      "grpc comes with a native extension that is built upon install. However, unless we happen ",
+      "to be deploying from the exact same architecture that we are deploying to, we need to ",
+      "recompile it for App Engine."
+    ],
+    "build-prod-grpc-extension": "cd ./server/node_modules/grpc && node_modules/.bin/node-pre-gyp install --fallback-to-build --library=static_library --target_arch=x64 --target_platform=linux --runtime=node --target=8.9.0 --target_libc=glibc",
```
gauntface: Will this need to change to `./node_modules` for grpc?

aomarks: Done.

package.json
aomarks: Done.

aomarks:
PTAL @samuelli and @gauntface!

This is updated for the big refactor in #67. It also moves the recompile-grpc and prune-node-deps steps into a new pre-deploy.js, for readability.

pre-deploy.js
```diff
@@ -0,0 +1,69 @@
+/**
```
samuelli: Why isn't this in TS? :)

pre-deploy.js
```diff
@@ -0,0 +1,69 @@
...
+const {promisify} = require('util');
+const execFile = promisify(require('child_process').execFile);
+const fse = require('fs-extra');
+const glob = promisify(require('glob'));
+
+/**
+ * grpc comes with a native extension that is built upon install. However,
+ * unless we happen to be deploying from the exact same architecture that we are
+ * deploying to, we need to recompile it for App Engine. This should be a
+ * non-issue once App Engine runs `npm install` for us.
```
samuelli: Can we check this is necessary still? Deploy should now be running `npm install` and will respect our `package-lock.json`.

pre-deploy.js
```diff
@@ -0,0 +1,69 @@
...
+  for (const grpcDir of grpcDirs) {
+    console.log(`rebuilding grpc extension at ${grpcDir}`);
+    await execFile('./node_modules/.bin/node-pre-gyp', args, {cwd: grpcDir});
+  }
+}
+
+/**
+ * We have a 20K file limit for App Engine. Our node_modules gets us close to
+ * this limit. Delete all unnecessary files to buy us some time. This should be
+ * a non-issue once App Engine runs `npm install` for us.
```
samuelli: This may also no longer be necessary.

PR #67, opened 2018-01-22
Refactor

gauntface:
Apologies for the size of this PR - it's a large number of files being moved and touched just to account for path changes.

With this PR, the CLI, server and client are moved to separate directories with shared logic pulled out into `types/` and `utils/`.

Everything under `src/` is copied to `build/` or built into `build/` following the same file structure.


package.json
```diff
@@ -1,32 +1,71 @@
...
-    "npm-install": "npm run npm-install:client && npm run npm-install:server",
-    "install": "npm run npm-install",
-    "deploy:gcloud": "gcloud app deploy app.yaml --project google.com:project-health-internal --quiet",
-    "deploy": "npm run build && npm run clean && npm run deploy:gcloud && npm run npm-install",
-    "format": "find server/src client/*.ts -name '*.ts' | xargs clang-format --style=file -i",
-    "precommit": "npm run format",
-    "server": "npm run build && node server.js"
+    "generate-gql-types": "apollo-codegen generate \"src/**/*.ts\" --schema \"src/types/github-schema.json\" --target typescript --addTypename --output \"src/types/gql-types.ts\"",
+    "clean": "rm -rf ./build/",
+    "copy": "cd src && cp --parents **/*.{html,json} ../build && cp --parents **/**/*.js ../build && cp -r --parents server/test/replays/**/* ../build",
```
aomarks: `--parents` is not a valid `cp` flag for me. I think it's probably only available in the GNU version, not the BSD version that comes with macOS. Do it some other way?

aomarks: Call this something more descriptive? Copy what?

package.json
```diff
@@ -1,32 +1,71 @@
...
+    "test:record": "npm run build && RECORD=true npm run test:ava -- --timeout=1min",
+    "test:ava": "nyc --reporter=text --reporter=html ava build/*/test/**/*-test.js --verbose"
   },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/PolymerLabs/project-health.git"
   },
   "license": "Apache-2.0",
   "devDependencies": {
+    "@types/command-line-args": "^4.0.2",
```
aomarks: I think it's ok to put `@types` in `devDependencies` because this is not a library, as discussed at https://github.com/Microsoft/types-publisher/issues/81. (No action required)

package.json
gauntface: I just switched to a node script after it failed on travis.

gauntface:
@aomarks I ended up moving to a node script to copy over the assets. The `cp` was flakey locally for me as well (ran differently on terminal vs npm run script).

build.js
```diff
@@ -0,0 +1,43 @@
+/**
```
aomarks: Would love for this to be typescript :-D

aomarks: Needs a license

src/types/shared-types.d.ts
```diff
@@ -1,4 +1,4 @@
-import {PullRequestReviewState} from 'server/src/gql-types';
```
samuelli: Maybe api-types.d.ts?

PR #70, opened 2018-01-24
Adding settings panel and tidying up push logic

gauntface:
This is largely moving the old push logic into settings and tidying up quite a bit.

I pulled out lit-html largely because I lost sight of why it was needed. I'm happy to add it back in but for now I'm not sure it's needed.

When we are ok to use Firestore I'll move over the push API calls to just adding and removing directly from firestore.


src/client/settings/push-component.ts
```diff
@@ -0,0 +1,107 @@
...
+      state.subscription = await registration.pushManager.getSubscription();
+      console.log(registration.pushManager);
+    })(),
+    (async () => {
+      const permissionState = await (navigator as any).permissions.query({name: 'notifications'});
+      state.permissionBlocked = permissionState.state === 'denied';
+    })()
+  ]);
+
+  console.log(state);
```
samuelli: Remove?

src/client/settings/push-component.ts
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  state.isSupported = !(!navigator.serviceWorker || !('PushManager' in window));
+  
+  // Run async tasks in parallel
+  await Promise.all([
+    (async () => {
+      const registration = await getRegistration();
+      state.subscription = await registration.pushManager.getSubscription();
+      console.log(registration.pushManager);
```
samuelli: Remove?

src/client/styles/base.css
```diff
@@ -0,0 +1,4 @@
+body {
+    font-family: Helvetica, Arial, sans-serif;
+    font-size: 14pt;
+  }
```
samuelli: indentation looks off

src/client/settings/push-component.ts
```diff
@@ -0,0 +1,107 @@
...
+    await unsubscribeUser(state.subscription);
+  } else {
+    await subscribeUser();
+  }
+}
+
+async function updateUI() {
+  const state = await getState();
+  if (!state.isSupported) {
+    pushStatus.textContent = `[Not Supported]`;
```
samuelli: Let's use single quote where possible

src/client/settings/push-component.ts
```diff
@@ -0,0 +1,107 @@
...
+
+function start() {
+  const toggleElement = <HTMLInputElement> document.querySelector('.push-component__toggle');
+  const statusElement = document.querySelector('.push-component__status');
+
+  if (!toggleElement) {
+    throw new Error(`Unable to find toggle element.`);
+  }
+  if (!statusElement) {
+    throw new Error(`Unable to find status element.`);
```
samuelli: These two as well

PR #73, opened 2018-01-24
Starting webhooks

gauntface:
This PR does a few things:

- Adds placeholders for the webhook endpoint (`/api/webhook/`). Once this is published, we should be in a position to add the webhook and starting getting data. I'm not sure how to add replays for webhooks.
- Moves the push subscription API out into a separate file to follow the same pattern as the APIs.
- Pulls out the login into into a utility function `getLoginFromRequest()`. This removes duplication and allows us to add some logic to avoid hitting Github unnecessarily. 
- Minor type tidy up of DashSecrets


src/server/apis/push-subscription.ts
```diff
@@ -0,0 +1,49 @@
+import * as express from 'express';
+import * as bodyParser from 'body-parser';
+
+import {GitHub} from '../../utils/github';
+import {PushSubscriptionModel} from '../models/PushSubscriptionModel';
+import { getLoginFromRequest } from '../utils/login-from-request';
```
samuelli: Feel like clang would've changed the formatting here.

src/server/dash-server.ts
```diff
@@ -1,33 +1,32 @@
...
-import {ViewerLoginQuery, ViewerPullRequestsQuery} from '../types/gql-types';
-import {PushSubscriptionModel} from './models/PushSubscriptionModel';
+import {ViewerPullRequestsQuery} from '../types/gql-types';
+import {getRouter as getWebhookRouter} from './apis/webhook';
+import {getRouter as getPushSubRouter} from './apis/push-subscription';
+
+class DashSecrets {
+  GITHUB_CLIENT_ID: string;
+  GITHUB_CLIENT_SECRET: string;
+}
```
samuelli: Should use type DashSecrets = {} instead.

src/server/apis/push-subscription.ts
```diff
@@ -0,0 +1,49 @@
...
+    } catch (err) {
+      console.error(err);
+      response.status(500).send(`An unhandled error occured.`);
+    }
+  });
+
+  return pushSubscriptionRouter;
+}
+
+export {getRouter};
```
samuelli: Fine for now, but I wonder if these should be classes.

src/server/apis/webhook.ts
```diff
@@ -0,0 +1,44 @@
...
+          return;
+        case 'pull_request':
+          await webhookEvents.handlePullRequest(request, response);
+          return;
+        case 'pull_request_review':
+          await webhookEvents.handlePullRequestReview(request, response);
+          return;
+        default:
+          console.warn(`Unsupported event type received: ${eventName}`);
+          response.status(400).send(`Unsupported event type.`);
```
samuelli: I assume we'll want to just add * for our webhooks.

For these events, we should respond 20x to GitHub. So here, I would use 202 to say "got it, did nothing".

PR #81, opened 2018-01-27
Show status strings based on PR information

samuelli:
This doesn't cover all the different scenarios that will be encountered. Each scenario should have a separate test case covering it.


src/client/dash.ts
```diff
@@ -50,7 +50,66 @@ function eventTemplate(event: PullRequestEvent) {
...
+      text = 'Pending your approval';
+      actionable = true;
+      break;
+    case PullRequestStatus.MergeRequired:
+      text = 'Requires merging';
+      actionable = true;
+      break;
+    default:
+      const unknown: never = pr.status;
+      throw new Error(`Unknown thing: ${unknown}`);
```
aomarks: Unknown PullRequestStatus

src/client/dash.ts
```diff
@@ -50,7 +50,66 @@ function eventTemplate(event: PullRequestEvent) {
...
+      text = 'New activity';
+      break;
+    case PullRequestStatus.StatusChecksPending:
+      text = 'Status checks pending';
+      break;
+    case PullRequestStatus.WaitingReview:
+      text = 'Waiting on reviewers';
+      break;
+    case PullRequestStatus.PendingChanges:
+      text = 'Waiting on you';
```
aomarks: I think I'd just do e.g. `return {text: 'Waiting on you', actionable: false};` in all of these cases, seems a bit simpler, and also makes it harder to accidentally forget a `break`.

src/server/dash-server.ts
```diff
@@ -34,7 +31,7 @@ export class DashServer {
     app.use('/node_modules/lit-html', express.static(litPath));
     app.use(express.static(path.join(__dirname, '../client')));
 
-    app.get('/dash.json', this.handleDashJson.bind(this));
+    app.get('/dash.json', new DashData(this.github).getHandler());
```
aomarks: Nice, I like the idea of using this pattern to separate every handler into its own file.

src/client/dash.ts
samuelli: Good point. I forgot to change it to that cause I had it inlined before instead of a function.

Done.

src/client/dash.ts
samuelli: Done

PR #85, opened 2018-01-29
Org hooks

gauntface:
This is a stepping stone to setting up the webhooks for orgs. Has issues with lit-html (can't disable checkbox - https://github.com/Polymer/lit-html/issues/257) and the next step will be to setup the webhooks.

### DIFF
--- a/src/client/oauth.html
+++ b/src/client/oauth.html
@@ -26,7 +26,10 @@
     } else if (queryParams.has('error')) {
       throw new Error(`Error returned by Github: '${queryParams.get('error_description')}'`);
     } else {
-      window.location = `https://github.com/login/oauth/authorize?client_id=${CLIENT_ID}&redirect_uri=https://project-health-internal.googleplex.com/oauth.html?redirect-origin=${REDIRECT_ORIGIN}`;
+      const scopes = [
+        'read:org',
+      ];
+      window.location = `https://github.com/login/oauth/authorize?client_id=${CLIENT_ID}&scope=${scopes.join(' ')},redirect_uri=https://project-health-internal.googleplex.com/oauth.html?redirect-origin=${REDIRECT_ORIGIN}`;
     }
   })();
 </script>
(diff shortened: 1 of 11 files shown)


src/server/apis/settings.ts
```diff
@@ -0,0 +1,96 @@
...
+    viewer {
+      organizations(first: 20) {
+        nodes {
+          name
+          viewerCanAdminister
+        }
+        totalCount
+      },
+    }
+  }`;
```
samuelli: Probably a reasonable assumption to assume they have less than 20 orgs. We can also change this a cursored query as well to ensure its always correct. All that's needed is to add the cursor information and call `GitHub.cursorQuery()`.

src/client/oauth.html
```diff
@@ -26,7 +26,10 @@
     } else if (queryParams.has('error')) {
       throw new Error(`Error returned by Github: '${queryParams.get('error_description')}'`);
     } else {
-      window.location = `https://github.com/login/oauth/authorize?client_id=${CLIENT_ID}&redirect_uri=https://project-health-internal.googleplex.com/oauth.html?redirect-origin=${REDIRECT_ORIGIN}`;
+      const scopes = [
+        'read:org',
```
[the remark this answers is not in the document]
gauntface: Moved to optional for now.

src/server/apis/settings.ts
gauntface: I've updated the TODO to with the cursorQuery.

I think it'll do for now, but it would be good to switch to cursor.

src/client/oauth.html
```diff
@@ -26,7 +26,11 @@
     } else if (queryParams.has('error')) {
       throw new Error(`Error returned by Github: '${queryParams.get('error_description')}'`);
     } else {
-      window.location = `https://github.com/login/oauth/authorize?client_id=${CLIENT_ID}&redirect_uri=https://project-health-internal.googleplex.com/oauth.html?redirect-origin=${REDIRECT_ORIGIN}`;
+      let scopes = queryParams.get('scopes');
```
samuelli: `let scopes = queryParams.get('scopes') || '';`

src/server/apis/settings.ts
samuelli: SGTM

PR #87, opened 2018-01-29
Sort outgoing PRs and add more tests

samuelli:


src/server/apis/dash-data.ts
```diff
@@ -159,7 +161,13 @@ export class DashData {
         incomingPrs.push(reviewedPr);
       }
     }
-    return {outgoingPrs, incomingPrs};
+    return {
+      // Sort newest first.
+      outgoingPrs: outgoingPrs.sort((a, b) => {
+        return a.createdAt < b.createdAt ? 1 : -1;
```
aomarks: Shouldn't this return 0 when they are equal? Could also do `outgoingPrs.sort((a, b) => b.createdAt - a.createdAt)`

samuelli: Done.

PR #89, opened 2018-01-30
Convert PullRequestStatus to a union of types

samuelli:


src/types/api.ts
```diff
@@ -0,0 +1,90 @@
...
+  reviewers: string[];
+}
+
+// Actionable - outgoing
+
+// Viewer's pull PR requires changes
+interface PendingChanges {
+  type: 'PendingChanges';
+}
+// Merge required by viewer
```
aomarks: add some newlines between these interfaces?

src/types/api.ts
```diff
@@ -0,0 +1,90 @@
...
+export interface Review {
+  author: string;
+  createdAt: number;
+  reviewState: PullRequestReviewState;
+}
+
+interface UnknownStatus {
+  type: 'UnknownStatus';
+}
+// Not necessarily actionable.
```
aomarks: move comment up

src/types/api.ts
samuelli: Done

src/types/api.ts
samuelli: Done.

PR #90, opened 2018-01-30
Settings with correct scope

gauntface:
- oauth.html can now include scopes + final-redirect so settings can request extra permissions
- Removed fake data and added a button to request permissions.


src/client/settings/org-webhook-list.ts
```diff
@@ -28,22 +28,35 @@ async function getState() {
...
 function hookTemplate(org: OrgWebHookState) {
-  // TODO: No idea how to add disabled state (i.e. bring
-  // org.viewerCanAdminister to checkbox)
+  const checkboxClick = () => {
+    console.log('Checkbox Click: ', org);
+  };
   return html`
   <div class="settings-toggle-item">
-    <input class="settings-toggle-item__toggle" type="checkbox"></input>
+    <input class="settings-toggle-item__toggle" type="checkbox" on-click="${() => checkboxClick()}"></input>
```
samuelli: Side comment: I wonder if we want to use a button like "Set up" instead of a toggle. Not sure we really need to support turning it off on our end - they can uninstall via GitHub.

src/client/settings/org-webhook-list.ts
```diff
@@ -28,22 +28,35 @@ async function getState() {
...
       <p>TODO: Add state message</p>
     </div>
   </div>`;
 }
 
+function requestPermissionTemplate() {
+  const readOrgsClick = () => {
+    window.location.href = `/oauth.html?scope=read:org&final-redirect=${window.location.href}`;
+  };
+  const requestPermissionTemplate = html`<p><button on-click="${() => readOrgsClick()}">Enable Org Reading</button></p>`;
```
samuelli: "Allow Project Health read-only access to your organizations"

src/client/settings/org-webhook-list.ts
```diff
@@ -28,22 +28,35 @@ async function getState() {
...
+  return requestPermissionTemplate;
+}
+
 async function updateUI() {
-  const state = await getState();
-  const orgTemplate = html`${state.orgs.map(hookTemplate)}`;
-  render(orgTemplate, orgListContainer);
+  try {
+    const state = await getState();
```
samuelli: Why is this a try/catch? Does  `getState()` throw when it doesn't have permission to read?

src/server/apis/settings.ts
```diff
@@ -8,71 +8,32 @@ import {OrgDetailsQuery} from '../../types/gql-types';
...
-                  }
-                ]
-              }
-            }
-          }
-        }
-      }
 
-      const orgs = orgDetails.data.viewer.organizations.nodes;
+      const orgDetails = await github.query<OrgDetailsQuery>({
```
samuelli: We should perhaps store the granted scopes by token when the user logs in. Maybe file an issue for this? Or a TODO somewhere?

src/client/settings/org-webhook-list.ts
gauntface: Yes - if the API can't request the orgs from github, our server returns a 500, which throws the fetch  in `getState` . I can catch in getState() and return null or return an object with `readOrgGranted: false` and true otherwise.

This decision was made along the lines of - this is a bad / undesirable situation to be in.

src/server/apis/settings.ts
gauntface: There is the `X-OAuth-Scopes` but I can't see a clean place at the moment to pull it from the login / graphql stuff.

Github Info: https://developer.github.com/apps/building-oauth-apps/scopes-for-oauth-apps/

src/client/settings/org-webhook-list.ts
gauntface: This is toggling the org webhook so we could just have setup and change - I'm just copying the mocks at the moment and personally I'm a fan of - "The place you set up is the place you turn it off". Setting up and then leaving it is kinda of a pain when I've used sites that do it.

PR #94, opened 2018-01-30
Add dashboard param to view other logins using your own token

samuelli:


src/server/apis/dash-data.ts
```diff
@@ -24,8 +24,9 @@ export class DashData {
       return;
     }
 
+    const username = req.query.login !== 'null' ? req.query.login : loginDetails.username;
```
aomarks: Looks like `null` is actually a valid user: https://github.com/null 

Use empty string?

samuelli: Oh, I just realised why I was getting `"null"`. Fixed and fixed.

PR #95, opened 2018-01-31
Filter reviews and PRs by the viewer

samuelli:
If the viewer created a PR and responded to review comments, this PR would show up in the incoming list as reviewed and these reviews would show up in the outgoing section. This change prevents removes these PRs from the incoming list and removes the reviews from the viewer.

### DIFF
--- a/src/server/apis/dash-data.ts
+++ b/src/server/apis/dash-data.ts
@@ -76,8 +76,9 @@ export class DashData {
 
         if (pr.reviews && pr.reviews.nodes) {
           // Filter out reviews from the viewer.
-          const prReviews = pr.reviews.nodes.filter((review) => review && review.author &&
-                                      review.author.login !== login);
+          const prReviews = pr.reviews.nodes.filter(
+              (review) =>
+                  review && review.author && review.author.login !== login);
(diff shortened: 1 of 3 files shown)


src/server/apis/dash-data.ts
```diff
@@ -87,10 +88,10 @@ export class DashData {
             reviewersCount === 0) {
           outgoingPr.status = {type: 'NoReviewers'};
         } else if (outgoingPr.status.type === 'WaitingReview') {
-          outgoingPr.status.reviewers = [
+          outgoingPr.status.reviewers = unique([
```
aomarks: Could just do:

```ts
outgoingPr.status.reviewers = Array.from(new Set([
  ...outgoingPr.reviewRequests,
  ...outgoingPr.reviews.map((review) => review.author),
]))
```

aomarks: This also works:

```ts
outgoingPr.status.reviewers = [...new Set([
  ...outgoingPr.reviewRequests,
  ...outgoingPr.reviews.map((review) => review.author),
])]
```

samuelli: Very cool. I went with the frist since I think its little bit clearer.

PR #97, opened 2018-01-31
Webhook adding

gauntface:
This isn't a complete feature but after getting the the API calls working, figured I'd stop before the PR grew too large.

- Scopes are added to cookies along side the id cookie.
- Scopes now seem happy and working
- Adding webhooks on a checkbox click is working.

After this - I'll need to:
1. Set initial state of the web hook.
1. Add support to remove the hook.

Both of these require storing the Hook ID


src/client/settings/org-webhook-list.ts
```diff
@@ -28,22 +29,65 @@ async function getState() {
...
+      body: JSON.stringify({
+        org: org.login,
+      }),
+      headers: {
+        'Content-Type': 'application/json',
+      }
+    });
+
+    if (!response.ok) {
+      // This the response is not 200, revert the checkbox state.
```
samuelli: Fix comment wording.

src/client/settings/org-webhook-list.ts
```diff
@@ -28,22 +29,65 @@ async function getState() {
...
   };
+
+  let message = `You cannot enable updates for ${org.name}. Request an owner to add the project-health WebHook.`;
+  if (org.viewerCanAdminister) {
+    // TODO: Handle scenario where the webhook is added.
+    message = `Toggle to enable updates for ${org.name}.`;
+  }
+
+  // TODO: Test if org has rights and if not, redirect user to:
+  // https://github.com/settings/connections/applications/:client_id ???
```
samuelli: What will that do?

src/server/apis/webhook.ts
```diff
@@ -38,7 +41,87 @@ function getRouter(): express.Router {
...
+        return;
+    }
+
+    if (!request.body.org) {
+      response.status(400).send('No org provided.');
+      return;
+    }
+
+    if (request.params.action === 'add') {
+      const githubAPIResponse = await fetch(
```
samuelli: Can we refactor this into github.ts? There's a v3 `get` API call, so a similar one can be added for `post`.  This will factor out some of the boilerplate like setting tokens, user agent, headers etc.

src/server/apis/webhook.ts
```diff
@@ -38,7 +41,87 @@ function getRouter(): express.Router {
...
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify({
+            name: 'web',
+            active: true,
+            events: [
+              '*',
+            ],
+            config: {
+              url: 'https://project-health-internal.googleplex.com/api/webhook',
```
samuelli: What's our plan here?

src/server/apis/settings.ts
```diff
@@ -14,6 +14,13 @@ function getRouter(github: GitHub): express.Router {
         return;
     }
 
+    const scopes = request.cookies['scope'] ?
+      request.cookies['scope'].split(',') : [];
+    if ((scopes.indexOf('admin:org_hook') === -1 || scopes.indexOf('read:org') === -1)) {
+      response.status(400).send('Missing required scope.');
+      return;
+    }
```
samuelli: Can we instead store the scopes in the user model and perform the check there? Doesn't seem robust to check the cookie here.

gauntface: Done with a chunk of other tidy up - sorry.

src/server/apis/webhook.ts
gauntface: Added - didn't spot that before.

src/server/apis/webhook.ts
gauntface: Moving to public - will keep as this for now just to carry on work and will have to change at a later stage.

gauntface:
@samuelli hopefully this is ready for round two of reviews.

src/server/apis/webhook.ts
```diff
@@ -38,7 +40,83 @@ function getRouter(): express.Router {
...
+                '*',
+              ],
+              config: {
+                url: 'https://project-health-internal.googleplex.com/api/webhook',
+                content_type: 'json'
+              }
+          });
+
+        const webhookId = githubResponseBody.id;
+        console.log(`Must add webhook '${webhookId}' to webhook ` +
```
samuelli: Why do we need the webhook id later? I'm not sure if we should be storing state related to the webhooks. To generate the settings page, we should read the orgs & webhooks directly. This is going to be easier than maintaining state since some you can remove it from GitHub settings too.

src/server/models/userModel.ts
```diff
@@ -0,0 +1,65 @@
...
+import {GitHub} from '../../utils/github';
+import {ViewerLoginQuery} from '../../types/gql-types';
+
+type LoginDetails = {
+  username: string;
+  token: string;
+  scopes: string[];
+}
+
+class UserModel {
```
samuelli: Thanks for doing this refactor!

src/server/test/models/PushSubscriptionModel-test.ts
```diff
@@ -21,44 +21,44 @@ const PUSH_SUBSCRIPTION_2 = {
 };
 const CONTENT_ENCODINGS = ['aes128gcm', 'aesgcm'];
 
-test('constructor', (t) => {
-  const model = new PushSubscriptionModel();
-  t.truthy(model);
+test.beforeEach(() => {
+  const subscriptions = pushSubscriptionModel.getSubscriptionsForUser(USER_LOGIN);
+  if (subscriptions) {
+    const subscriptionKeys =Object.keys(subscriptions);
```
samuelli: Format doesn't look like its committed.

src/types/api.ts
```diff
@@ -1,5 +1,9 @@
 import {PullRequestReviewState} from './gql-types';
 
+export type DashSecrets = {
+  GITHUB_CLIENT_ID: string; GITHUB_CLIENT_SECRET: string;
+};
```
samuelli: I don't think this should live here, `api.ts` is meant to document the types necessary to share between the server and client.

PR #98, opened 2018-01-31
Add new commits event

samuelli:
This change rolls changes to the GitHub schema and adds the new commits event for pull requests that have been reviewed.

image

### DIFF
--- a/src/server/apis/dash-data.ts
+++ b/src/server/apis/dash-data.ts
@@ -97,7 +97,7 @@ export class DashData {
         outgoingPrs.push(outgoingPr);
       }
 
-      // Incoming reviews
+      // Incoming review requests.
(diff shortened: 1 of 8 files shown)


src/server/apis/dash-data.ts
```diff
@@ -148,6 +149,37 @@ export class DashData {
           reviewedPr.events.push({type: 'MyReviewEvent', review: myReview});
         }
 
+        // Check if there are new commits to the pull request.
+        if (pr.commits.nodes) {
+          const newCommits = pr.commits.nodes.filter(
```
aomarks: Do this filter in the loop below so that you don't loop twice and don't parse the `pushedDate` twice? Fine if you prefer this, but seems simpler to me to do one loop.

samuelli: Fair point. Thanks, done.

PR #109, opened 2018-02-01
Add link to changes for new commits

samuelli:
For event like "2 new commits", this adds a link to view the diff in GitHub from the latest commit to the last time the viewer reviewed the PR.


samuelli:
I've found a complication around cases in which force push was used, eg when rebasing. GitHub won't be able to show the diff for these cases. Not entirely sure what to do here. We could check all the commits, but we have to be sure that we have all of them and this could result in an inconsistent experience. Happy to discard for now if that's better.

aomarks:
> I've found a complication around cases in which force push was used, eg when rebasing. GitHub won't be able to show the diff for these cases. Not entirely sure what to do here. We could check all the commits, but we have to be sure that we have all of them and this could result in an inconsistent experience. Happy to discard for now if that's better.

What does it look like for the user when this happened? Is there some way to know that this happened?

samuelli:
> What does it look like for the user when this happened? Is there some way to know that this happened?

Sample link: https://github.com/PolymerLabs/project-health/pull/97/files/c986964fb78eaed47b7a63cb2be69014a54fcbd2..1efc29ce79d1269115598330e96dc75fe1d2011c

GitHub displays the following:
image

aomarks:
Seems fine to me, it will work most of the time right?

PR #111, opened 2018-02-02
Webhooks

gauntface:
- Adds ability to remove hooks
- Sets initial state correctly.

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,7 +1,7 @@
 'use strict';
 
-const {DashServer} = require('./build/server/dash-server');
-const {GitHub} = require('./build/utils/github');
+const { DashServer } = require('./build/server/dash-server');
+const { GitHub } = require('./build/utils/github');
 const fs = require('fs');
 const path = require('path');
 const secrets = require('./secrets.json');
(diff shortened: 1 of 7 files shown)


server.js
```diff
@@ -4,5 +4,12 @@ const {DashServer} = require('./build/server/dash-server');
 const {GitHub} = require('./build/utils/github');
 const secrets = require('./secrets.json');
 
+const logError = (err) => {
+  console.error(err);
+  throw err;
+};
+process.on('uncaughtException', logError);
+process.on('unhandledRejection', logError);
```
samuelli: Can we move this to the end of the file?

src/server/apis/settings.ts
```diff
@@ -35,8 +34,34 @@ function getRouter(github: GitHub): express.Router {
...
+      const orgsData = orgDetails.data.viewer.organizations.nodes;
+      if (orgsData) {
+        for (const org of orgsData) {
+          if (!org) {
+            continue;
+          }
+  
+          let hookEnabled = false;
+          if (org.viewerCanAdminister) {
+            const hooks = await github.get(`orgs/${org.login}/hooks`, loginDetails.token);
```
samuelli: This executes in serial. Can we change it to parallel by using Promise.all instead of await?

server.js
gauntface: done

PR #112, opened 2018-02-02
Add favicon

samuelli:
This change doesn't change the favicon when there is new activity. For that we'll need to poll for new data and be able to keep the dashboard open for a long time. This will also require persistence of user credentials in firestore.


src/client/dash.html
```diff
@@ -3,9 +3,11 @@
 
 <head>
   <meta name="viewport" content="width=device-width,initial-scale=1">
-  <title>Action Dash</title>
+  <title>Project Health</title>
   <link href="/styles/base.css" rel="stylesheet" />
   <link href="/styles/dash.css" rel="stylesheet" />
+  <link rel="icon" type="image/png" href="/assets/favicon-32x32.png?v=2" sizes="32x32">
+  <link rel="icon" type="image/png" href="/assets/favicon-16x16.png?v=2" sizes="16x16">
```
gauntface: Is the expectation that the version will manually get bumped with changes?

samuelli: Yep. I should remove that for now, sorry.

PR #114, opened 2018-02-03
Push notifications

gauntface:
- SW needed to be pulled out into seperate file to work with typescript.
- Altered the push subscriptions model to keep an array of subscriptions per user. This needs to be moved to firestore.
- Tests are just plain smoke tests. I need to think of a way detecting a notification being sent with the right details.


src/client/settings/push-details.ts
```diff
@@ -1,5 +1,6 @@
+// TODO: Get from server side code or form secrets somehow.
```
samuelli: from?

src/server/controllers/notifications.ts
```diff
@@ -0,0 +1,23 @@
+/// <reference path="../../types/web-push.d.ts" />
+import * as webpush from 'web-push';
+
+import {pushSubscriptionModel} from '../models/pushSubscriptionModel';
+import {DashSecrets} from '../dash-server';
+import {NotificationPayload} from '../../types/api';
+
+export const sendNotification = (secrets: DashSecrets, recipient: string, data: NotificationPayload) => {
+  const userSubscriptionDetailss = pushSubscriptionModel.getSubscriptionsForUser(recipient);
```
samuelli: typo

src/client/settings/push-details.ts
samuelli: It could be injected when we serve this file.

src/server/test/apis/webhook-test.ts
```diff
@@ -0,0 +1,32 @@
...
+  GITHUB_CLIENT_SECRET: 'ClientSecret',
+  PUBLIC_VAPID_KEY: 'BPtJjYprRvU3TOb0tw3FrVbLww3bp7ssGjX99PFlqIOb3b8uOH4_Q21GYhwsDRwcfToaFVVeOxWOq5XaXD1MGdw',
+  PRIVATE_VAPID_KEY: 'o1P9aXm-QPZezF_8b7aQabivhv3QqaB0yg5zoFs6-qc',
+};
+
+test('Webhook pull_request_review: submitted-state-changes_requested.json', async (t) => {
+  const eventContent = await fs.readJSON(path.join(hookJsonDir, 'pull_request_review', 'submitted-state-changes_requested.json'));
+  await handlePullRequestReview(TEST_SECRETS, eventContent);
+  
+  // For now nothing to test - should just pass
```
samuelli: This should probably just test the boundary of the server. ie. check that the server calls `webpush.sendNotification()` with the correct params.

src/server/webhook-events.ts
```diff
@@ -1,26 +1,63 @@
...
+      notification = {
+        title: `${review.user.login} approved your PR`,
+        body: `[${repo.name}] ${pullReq.title}`,
+      };
+    } else if (review.state === 'changes_requested') {
+      notification = {
+        title: `${review.user.login} requested changes`,
+        body: `[${repo.name}] ${pullReq.title}`,
+      };
+    }
```
samuelli: Don't we also want commented?

src/sw/sw.ts
```diff
@@ -0,0 +1,22 @@
+import {} from '.';
+declare var self: ServiceWorkerGlobalScope;
+
+import {NotificationPayload} from '../types/api';
+
+// This isn't a function to make typescript happy
+// intead its an "EventListener"
```
samuelli: instead

samuelli: Probably should update the comment to says its just an EventListener

src/sw/sw.ts
```diff
@@ -0,0 +1,22 @@
...
+    if (!event.data) {
+      return;
+    }
+
+    const data: NotificationPayload =
+        (JSON.parse(event.data.text()) as NotificationPayload);
+    event.waitUntil(self.registration.showNotification(
+        data['title'],
+        data,
+        ));
```
samuelli: Out of curiosity, are we required to show a notification when we receive a push event? Or can we simply cause the dashboard to refresh?

src/sw/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "noUnusedParameters": true,
+    "pretty": true,
+    "skipLibCheck": true,
+    "lib": [
+      "es2017",
+      "esnext.asynciterable",
+      "webworker"
+    ]
+  }
+}
```
samuelli: @aomarks should this change at all based or your new src/tsconfig.json?

src/server/test/apis/webhook-test.ts
gauntface: Originally I wanted to just test if our code was called (i.e. the sendNotification() function) and ensure the parameters, but I can't see a way to do it. Typically I'd just use sinon to spy on the function and examine the call count and arguments.

That isn't working in this case, possible because of the typescript build?

Do you have any example of doing this?

src/sw/sw.ts
gauntface: We don't have to show a notification if there is a tab open and focused. In that scenario we can just refresh the pages (and we can refresh the pages regardless).

src/sw/tsconfig.json
gauntface: @aomarks let me know if you'd like this to change in any way.

aomarks: Make sure to add `sw` to the `exclude` section of the top-level `src/tsconfig.json`.

PR #115, opened 2018-02-03
Deploy from build/ and rewrite server launcher in TypeScript.

### LINKED ISSUE #101: Move npm install --production into build

### LINKED ISSUE #103: Convert server.js to ts

aomarks:
- We now deploy from `build/` instead of the root directory, and we no longer run `npm install` since App Engine now does it for us. Fixes #101.

- The script to launch the server is now in TypeScript at `src/server.ts`, so the compiled version is at `build/server.js`. Fixes #103.

- Moved `src/server/tsconfig.json` up a level, and got rid of the separate `cli/` build target. `src/tsconfig.json` now builds everything except `client/` (by excluding it explicitly). I did this because App Engine requires that our launcher is exactly at `build/server.js`, and this was the simplest way to do it.

- Renamed `npm run deploy` to `npm run deploy:staging` to be more clear.

- Renamed `build.js` to `copy-static-assets.js` to be more clear.


package.json
```diff
@@ -6,15 +6,13 @@
...
-    "deploy": "npm run build && rm -rf ./node_modules/ && npm install --only=production && npm run deploy:gcloud -- --project github-health-staging && npm install",
-    "deploy:prod": "npm run build && rm -rf ./node_modules/ && npm install --only=production && npm run deploy:gcloud -- --project github-health && npm install",
-    "deploy:gcloud": "gcloud app deploy app.yaml --quiet",
+    "build:server": "tsc -p src/tsconfig.json --outDir build",
+    "build": "npm run clean && npm run generate-gql-types && npm run build:client && npm run build:server && node copy-static-assets.js && cp {package.json,package-lock.json,app.yaml,secrets.json,github-health-*.json} build/ && npm run lint",
+    "build:watch": "npm run build && (npm run build:server -- --watch & npm run build:client -- --watch)",
+    "deploy:staging": "npm run build && cd build && gcloud app deploy app.yaml --quiet --project github-health-staging",
+    "deploy:prod": "npm run build && cd build && gcloud app deploy app.yaml --quiet --project github-health",
+    "launch": "npm run build && node build/server.js",
```
samuelli: `start` is usually more typical

aomarks: Hmm, GAE at some point will respect `start` right? In which case, this would be totally wrong. Maybe I'll just remove it for now.

aomarks:
Travis fixed, and successfully deployed to staging.

PR #121, opened 2018-02-06
Push Model -> Firestore

gauntface:
This moves a users push subscriptions over to Firestore.

### DIFF
--- a/src/server/models/pushSubscriptionModel.ts
+++ b/src/server/models/pushSubscriptionModel.ts
@@ -1,3 +1,5 @@
+import {Firestore} from '@google-cloud/firestore';
+
 interface PushSubscription {
   endpoint: string;
   expirationTime: number|null;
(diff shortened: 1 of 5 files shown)


src/server/models/pushSubscriptionModel.ts
```diff
@@ -10,48 +12,53 @@ interface PushSubscriptionInfo {
...
-  private pushSubscriptions: {[id: string]: PushSubscriptionInfo[]};
+  private firestore: Firestore;
 
   constructor() {
-    this.pushSubscriptions = {};
+    this.firestore = new Firestore();
+  }
+
+  _getSubscriptionCollection(login: string) {
+    const collectionPath = `pushsubscriptions/${login}/subscriptions`;
```
aomarks: How about we call the top-level collection `users`? Seems redundant to have `pushsubscriptions > $login > subscriptions`, but `users > $login > subscriptions` makes sense. Then we can shove other user info in the same user documents.

aomarks: How about `this.firestore.collection('users').doc(login).collection('subscriptions')`? Seems a bit more clear what's going on, and avoids any issue where `login` might have a `/secrets/` or something in the string.

src/server/models/pushSubscriptionModel.ts
```diff
@@ -10,48 +12,53 @@ interface PushSubscriptionInfo {
...
 class PushSubscriptionModel {
-  private pushSubscriptions: {[id: string]: PushSubscriptionInfo[]};
+  private firestore: Firestore;
 
   constructor() {
-    this.pushSubscriptions = {};
+    this.firestore = new Firestore();
+  }
+
+  _getSubscriptionCollection(login: string) {
```
aomarks: use `private` instead of underscore prefix

src/server/models/pushSubscriptionModel.ts
```diff
@@ -10,48 +12,53 @@ interface PushSubscriptionInfo {
...
+    const querySnapshot = await subCol.get();
+    return querySnapshot.docs.map((doc) => {
+      return (doc.data() as PushSubscriptionInfo);
+    });
   }
 }
 
-export const pushSubscriptionModel = new PushSubscriptionModel();
+let subscriptionModel: PushSubscriptionModel;
+export const getSubscriptionModel = () => {
```
aomarks: Why not `export function`?

src/server/test/models/PushSubscriptionModel-test.ts
```diff
@@ -21,92 +23,71 @@ const PUSH_SUBSCRIPTION_2 = {
...
-  const subscriptions =
-      pushSubscriptionModel.getSubscriptionsForUser(USER_LOGIN);
-  if (subscriptions) {
-    subscriptions.forEach((subscriptionInfo) => {
-      pushSubscriptionModel.removePushSubscription(
-          USER_LOGIN, subscriptionInfo.subscription);
-    });
+test.beforeEach(async () => {
+  // See https://cloud.google.com/docs/authentication/production
+  if (!process.env.GOOGLE_APPLICATION_CREDENTIALS) {
```
aomarks: Hmm, maybe we should just do this in dash-server or something. Fine for now though.

src/server/models/pushSubscriptionModel.ts
aomarks: Could you add a comment somewhere explaining the structure of the db (including the fact that it's keyed by base64 of endpoint etc.)

PR #127, opened 2018-02-06
Force HTTPS

### LINKED ISSUE #125: Enforce HTTPS usage

samuelli:
Fixes #125


src/server/dash-server.ts
```diff
@@ -76,6 +77,15 @@ export class DashServer {
     }
   }
 
+  enforceHTTPS(req: express.Request, res: express.Response, next: express.NextFunction) {
+    if (!req.secure && process.env.NODE_ENV === 'production') {
+      res.header('Location', `https://${req.headers['host']}${req.url}`);
```
aomarks: you can use `res.redirect(code, url)`

src/server/dash-server.ts
```diff
@@ -76,6 +77,15 @@ export class DashServer {
...
+      res.sendStatus(301);
+      return;
```
aomarks: use else instead of return

src/server/dash-server.ts
samuelli: Oh cool, thanks.

src/server/dash-server.ts
samuelli: Done.

PR #130, opened 2018-02-07
Gulp

### LINKED ISSUE #102: Use gulp for our build process

gauntface:
- `gulp --tasks` will show the available tasks. 
- `gulp build` will clean the build directory, gen GQL types, lint then build the typescript, copy and minify CSS, images, html and copy over json files in parallel.
- `gulp dev` will run the `build` task, then start the server + run the `watch` task, this will rebuild any files as they change and restart the server after that.

This PR also adds `npm run start` which is working on staging at the moment.

Fixes #102


gulp-tasks/build.js
```diff
@@ -0,0 +1,33 @@
...
+    if (build) {
+      buildTasks.push(build);
+    }
+  }
+
+  return gulp.series([
+    cleanDestDir,
+    genGQLTypes,
+    lint,
+    gulp.parallel(buildTasks),
```
samuelli: Cool!

gulp-tasks/copy.js
```diff
@@ -0,0 +1,22 @@
...
+
+const extensions = [
+  'json',
+];
+
+const extensionsGlob = () => {
+  if (extensions.length === 1) {
+    return `${global.__buildConfig.src}/**/*.${extensions[0]}`
+  } else {
+    return `${global.__buildConfig.src}/**/*.{${extensions.join(',')}}`
```
samuelli: `.{json}` doesn't work? If it does, if/else isn't needed.

gulp-tasks/copy.js
```diff
@@ -0,0 +1,22 @@
+const gulp = require('gulp');
+
+const extensions = [
+  'json',
+];
+
+const extensionsGlob = () => {
+  if (extensions.length === 1) {
+    return `${global.__buildConfig.src}/**/*.${extensions[0]}`
```
samuelli: `;` here and below.

gulp-tasks/css.js
```diff
@@ -0,0 +1,30 @@
+const gulp = require('gulp');
+
+const postcss = require('gulp-postcss');
+const cssimport = require('postcss-import');
+const cssnext = require('postcss-cssnext');
+const cssnano = require('cssnano');
```
samuelli: I'm unfamiliar with all of these. If you know roughly what each is used for, can you add a quick comment? eg. "This is used to minify the CSS"

package.json
```diff
@@ -4,20 +4,15 @@
   "description": "A tool to help Polymer team manage repos.",
   "private": true,
   "scripts": {
+    "start": "node ./build/server.js",
```
samuelli: I started sorting these alphabetically, might be nice to maintain that. :)

package.json
```diff
@@ -4,20 +4,15 @@
...
     "generate-gql-types": "apollo-codegen generate 'src/**/*.ts' --schema 'src/types/github-schema.json' --target typescript --addTypename --output 'src/types/gql-types.ts'",
-    "clean": "rm -rf ./build/",
-    "build:client": "tsc -p src/client/tsconfig.json --outDir build",
-    "build:server": "tsc -p src/tsconfig.json --outDir build",
-    "build:sw": "tsc -p src/sw/tsconfig.json --outDir build",
-    "build": "npm run clean && npm run generate-gql-types && npm run build:client && npm run build:sw && npm run build:server && node copy-static-assets.js && cp package.json package-lock.json app.yaml build/ && cp secrets.json build/ | true && cp github-health-*.json build/ | true && npm run lint",
```
samuelli: Should build still point to gulp build? Likewise for dev?

src/server/test/apis/webhook-test.ts
```diff
@@ -17,7 +17,7 @@ const TEST_SECRETS = {
 test.before(() => {
   // See https://cloud.google.com/docs/authentication/production
   if (!process.env.GOOGLE_APPLICATION_CREDENTIALS) {
-    const rootDir = path.join(__dirname, '..', '..', '..');
+    const rootDir = path.join(__dirname, '..', '..', '..', '..');
```
samuelli: `path.resolve(__dirname, '../../../../')` seems clearer to me.

gulp-tasks/build.js
```diff
@@ -0,0 +1,33 @@
+const gulp = require('gulp');
+const fse = require('fs-extra');
+
+const getTaskFilepaths = require('./utils/get-task-filepaths');
+const {genGQLTypes} = require('./utils/genGQLTypes');
```
samuelli: `genGQLTypes` isn't used anywhere else. Can we inline that file here?

gulp-tasks/build.js
```diff
@@ -0,0 +1,33 @@
...
+const {lint} = require('./utils/lint');
```
samuelli: Likewise, `lint` is only used here. Inline?

gulp-tasks/utils/npmRunScript.js
```diff
@@ -0,0 +1,13 @@
+const path = require('path');
+const spawn = require('./spawn-promise');
+
+const npmRunScript = (scriptName) => {
```
samuelli: If `lint` and `genGQLTypes` get inlined, this could also get folded into `build`, or at least `spawn-promise` gets inlined here.

src/server/test/apis/webhook-test.ts
aomarks: Double check if that works on windows though?

gulp-tasks/copy.js
gauntface: It doesn't work :(

gulp-tasks/utils/npmRunScript.js
gauntface: Generally, there are useful enough that I reckon we'll end up wanting to re-use these.

src/server/test/apis/webhook-test.ts
gauntface: It won't work on Windows - hence this.

PR #134, opened 2018-02-07
Fixing the uncached user issue

gauntface:
When the user has a token but the server doesn't have the details cached, they have to re-login, this PR changes that so the existing token is used - although we do lose the scope details.

### DIFF
--- a/src/server/apis/dash-data.ts
+++ b/src/server/apis/dash-data.ts
@@ -18,7 +18,7 @@ export class DashData {
   }
 
   private async handler(req: express.Request, res: express.Response) {
-    const loginDetails = await userModel.getLoginFromRequest(req);
+    const loginDetails = await userModel.getLoginFromRequest(this.github, req);
     if (!loginDetails) {
       res.sendStatus(401);
       return;
--- a/src/server/apis/settings.ts
+++ b/src/server/apis/settings.ts
@@ -13,13 +13,18 @@ function getRouter(github: GitHub): express.Router {
   settingsRouter.post(
       '/orgs.json',
       async (request: express.Request, response: express.Response) => {
-        const loginDetails = await userModel.getLoginFromRequest(request);
+        const loginDetails =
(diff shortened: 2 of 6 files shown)


src/server/apis/dash-data.ts
```diff
@@ -18,7 +18,7 @@ export class DashData {
   }
 
   private async handler(req: express.Request, res: express.Response) {
-    const loginDetails = await userModel.getLoginFromRequest(req);
+    const loginDetails = await userModel.getLoginFromRequest(this.github, req);
```
samuelli: We should probably make that singleton :)

src/server/models/userModel.ts
```diff
@@ -7,7 +7,7 @@ import {GitHub} from '../../utils/github';
 interface LoginDetails {
   username: string;
   token: string;
-  scopes: string[];
+  scopes: string[]|null;
```
samuelli: Can't we just keep it as is? [] can the empty list.

src/server/apis/settings.ts
```diff
@@ -13,13 +13,18 @@ function getRouter(github: GitHub): express.Router {
...
-        const loginDetails = await userModel.getLoginFromRequest(request);
+        const loginDetails =
+            await userModel.getLoginFromRequest(github, request);
         if (!loginDetails) {
           response.sendStatus(400);
           return;
         }
 
         const scopes = loginDetails.scopes;
+        if (!scopes) {
```
samuelli: This doesn't return a response. If `null` is removed, this line won't be necessary, and it'll fall into the below 400 statement of missing scope.

src/server/models/userModel.ts
gauntface: The reason for this is so that we can differentiate between - we know the scopes and there are none granted and when we simply don't know the scopes.

This is likely to change when we move to Firestore, so for now I'd like to keep this as is.

PR #135, opened 2018-02-07
Adding support for hook status notifications

gauntface:
Adds support for notifications on status hook events.

Next PR will be adding the Application singleton (Which I keep whining about).

### DIFF
--- a/src/server/dash-server.ts
+++ b/src/server/dash-server.ts
@@ -12,6 +12,7 @@ import {getRouter as getLoginRouter} from './apis/login';
 import {getRouter as getPushSubRouter} from './apis/push-subscription';
 import {getRouter as getSettingsRouter} from './apis/settings';
 import {getRouter as getWebhookRouter} from './apis/webhook';
+import {enforceHTTPS} from './utils/enforce-https';
 
 export type DashSecrets = {
   GITHUB_CLIENT_ID: string; GITHUB_CLIENT_SECRET: string;
@@ -32,6 +33,10 @@ export class DashServer {
 
     const app = express();
     const litPath = path.dirname(require.resolve('lit-html'));
+
+    if (process.env.NODE_ENV === 'production') {
+      app.use(enforceHTTPS);
+    }
 
     app.use(cookieParser());
     app.use('/node_modules/lit-html', express.static(litPath));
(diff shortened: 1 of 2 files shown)


src/server/webhook-events.ts
```diff
@@ -1,29 +1,77 @@
...
+    // We have no details so can't get the PR for the commit
+    return;
+  }
+
+  const prData = await github.query<StatusToPRQuery>({
+    query: statusToPRQuery,
+    variables: {
+      queryString: `type:pr repo:${repo.full_name} ${hookData.sha}`,
+    },
+    context: {token: loginDetails.token}
```
samuelli: Useful to comment explicitly that we are using the PR author's token here.

src/server/webhook-events.ts
```diff
@@ -1,29 +1,77 @@
...
+  });
+  const nodes = prData.data.search.nodes;
+  if (nodes && nodes.length > 0) {
```
samuelli: Is it possible this SHA belongs to multiple PRs? eg. when I chain changes, and they all contain the same commit, but are in separate PRs to merge to master.

src/server/webhook-events.ts
```diff
@@ -66,3 +114,15 @@ export async function handlePullRequestReview(
         `Unexpected Pull Request Review Action: ${hookData.action}`);
   }
 }
+
+const statusToPRQuery = gql`
+query StatusToPR($queryString: String!) {
+  search(type: ISSUE, query:$queryString, first: 1) {
+    nodes {
+      ...on PullRequest {
+        id, number, title
```
samuelli: nit: can we newline separate these please? That's the style we've been using elsewhere.

src/server/webhook-events.ts
gauntface: It is possible - not sure of the scenario of this and what the expected behavior should be.

We could alter the message from `[<repo name>] <pull request name>` to `[<repo name>] <X PR Count>'s are affected by this status.`

gauntface:
Note to self: Get the latest commit for PR's the commit is in and only send notifications for PR's where this commit is the latest.

src/server/webhook-events.ts
samuelli: FTR, we discussed offline. This should probably fetch the last commits of the matching pull requests, and send a notification for each pull request that has this commit as the latest commit. Otherwise, we would be spamming on old commits for the PR.

PR #138, opened 2018-02-07
Add @mentioned events to incoming pull requests

### LINKED ISSUE #99: @mentioned events on incoming reviews

samuelli:
Fixes #99


src/server/apis/dash-data.ts
```diff
@@ -32,14 +32,16 @@ export class DashData {
...
 
     const viewerPrsResult = await this.github.query<ViewerPullRequestsQuery>({
       query: prsQuery,
       variables: {
         login,
         reviewRequestsQueryString: `review-requested:${login} ${openPrQuery}`,
-        reviewedQueryString:
-            `reviewed-by:${login} ${openPrQuery} -author:${login}`,
+        reviewedQueryString,
+        mentionsQueryString: `${reviewedQueryString} mentions:${login}`,
```
gauntface: would be nice if this followed the same pattern and `reviewedQueryString` and gets defined outside of this object.

src/server/apis/dash-data.ts
```diff
@@ -189,8 +199,24 @@ export class DashData {
           }
         }
 
+        prIds.push(pr.id);
         incomingPrs.push(reviewedPr);
       }
+
+      // Incoming review requests.
+      for (const pr of viewerPrsResult.data.reviewRequests.nodes || []) {
+        if (!pr || pr.__typename !== 'PullRequest' || prIds.includes(pr.id)) {
```
gauntface: The naming here isn't clear `prIds.includes(pr.id)`, how do these ID's differ?

src/server/apis/dash-data.ts
samuelli: Done

src/server/apis/dash-data.ts
samuelli: Updated name to prsShown. I just stumbled upon a weird case where GitHub shows it in review requests and there is a review on it. So this prevents it from showing up twice.

PR #139, opened 2018-02-07
Moves secrets, github and firestore into an application singleton

gauntface:
I'm not 100% convinced this is the neatest way to manage these values since everything is null by default, and getters are used at runtime to ensure the values have been set. The advantage is tests can set what is needed in terms of dependencies and leave everything else as null.

### DIFF
--- a/src/test/server/apis/dash-data-test.ts
+++ b/src/test/server/apis/dash-data-test.ts
@@ -3,6 +3,7 @@ import * as ava from 'ava';
 import {startTestReplayServer} from '../../../replay-server';
 import {DashData} from '../../../server/apis/dash-data';
 import {PullRequestReviewState} from '../../../types/gql-types';
+import {initGithub} from '../../../utils/github';
 
 /**
  * Assigns the test context object before each test to ensure it is correctly
@@ -20,11 +21,12 @@ function contextualize<T>(getContext: (_: ava.TestContext) => Promise<T>):
  * Generates the test context object before each test.
  */
 const test = contextualize(async (t) => {
-  const {server, client} = await startTestReplayServer(t);
+  const {server, url} = await startTestReplayServer(t);
+  initGithub(url, url);
+  
   return {
     replayServer: server,
-    client,
-    dashData: new DashData(client),
+    dashData: new DashData(),
     // This token must be set in the environment during recording.
     token: process.env.GITHUB_TOKEN || '',
   };
--- a/src/test/server/util/my-repos-test.ts
+++ b/src/test/server/util/my-repos-test.ts
@@ -2,6 +2,7 @@ import * as ava from 'ava';
 
 import {startTestReplayServer} from '../../../replay-server';
 import {getMyRepos} from '../../../server/utils/my-repos';
+import {initGithub} from '../../../utils/github';
 
 /**
  * Assigns the test context object before each test to ensure it is correctly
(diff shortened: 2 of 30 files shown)


src/application.ts
```diff
@@ -0,0 +1,55 @@
...
+  PUBLIC_VAPID_KEY: string;
+  PRIVATE_VAPID_KEY: string;
+};
+
+type Init = {
+  secrets: ApplicationSecrets|null; firestore: Firestore | null;
+  github: GitHub | null;
+};
+
+class Application {
```
aomarks: Does this need to be a class? If we're doing globals, maybe we could just have a module called "environment" that exports a bunch of globals. I find the name Application a little confusing, since it overlaps conceptually with the server class itself.

gauntface:
@aomarks could you double check that the changes to the `contextualize` steps aren't going to mess with the tests. Looking at them I couldn't tell how or why they were different to setting the context on the ava test.

samuelli:
> @aomarks could you double check that the changes to the contextualize steps aren't going to mess with the tests. Looking at them I couldn't tell how or why they were different to setting the context on the ava test.

The contextualise step is simply setting the context on the ava test. It's entire purpose is to provide accurate typings for everything in the context object. By default, the context object is `any`, which isn't very useful.

src/utils/firestore.ts
```diff
@@ -0,0 +1,25 @@
...
+let firestoreSingleton: Firestore|null = null;
+
+export function firestore(): Firestore {
+  if (!firestoreSingleton) {
+    throw new Error('Firestore is not initialised.');
+  }
+  return firestoreSingleton;
+}
+
+export function initFirestore(keyFile: string|void) {
```
samuelli: It looks like all calls to `initFirestore()` are almost identical. Can we instead move all the logic to `firestore()`?

If it hasn't been initialised, we initialise the singleton. The only difference I could see was that in tests, we do not want to use env variable credentials. We can easily check which case we are in by checking `process.env.NODE_ENV` to check if its a testing environment.

This will mean that firestore knows how to initialise itself for different environments and users of it simply need to simply call `firestore()`

src/test/server/util/my-repos-test.ts
```diff
@@ -1,44 +1,22 @@
...
-/**
- * Generates the test context object before each test.
- */
-const test = contextualize(async (t) => {
-  const {server, client} = await startTestReplayServer(t);
-  return {
-    server,
-    client,
-  };
-});
```
samuelli: This needs to be brought back.

src/test/server/apis/dash-data-test.ts
```diff
@@ -1,33 +1,19 @@
...
- */
-const test = contextualize(async (t) => {
-  const {server, client} = await startTestReplayServer(t);
-  return {
-    replayServer: server,
-    client,
-    dashData: new DashData(client),
-    // This token must be set in the environment during recording.
-    token: process.env.GITHUB_TOKEN || '',
-  };
```
samuelli: This also needs to be brought ba ck.

src/utils/firestore.ts
gauntface: -1 this approach.

I would prefer an API where we force a conscious initialisation and we make it clear when that init is valid and applied vs when it's not.

(i.e. if I call firestore() and then I call firestore('/some-key-file.json') should it use the key file or not?)

gauntface:
@samuelli I re-added the contextualize logic. I'm not a fan of the firestore proposal - if you really want it, I'll add it back in but I think it's a bad precedence to set.

src/utils/firestore.ts
samuelli: There's 2 ideas I conflated in there. One was suggesting merging `firestore()` and `initFirestore()` which I understand your objections.

The second is to move all the logic into `initFirestore()`. This would have all the logic in `initFirestore()` (no params), and it'll check for the `test` env, and decide what credentials to load. It wouldn't allow for another key file, but I don't see that being necessary and this will allow the rest of the code to simply call:
```
initFirestore()
firestore()
```

PR #141, opened 2018-02-08
Message user on breaking status and fix commenting your own PR

gauntface:


src/server/controllers/webhook-events.ts
```diff
@@ -21,8 +26,76 @@ type PullRequestReviewHook = {
   repository: RepositoryHook;
 };
 
+type StatusHook = {
+  sha: string;
+  name: string;
+  state: 'error'|'failure'|'pending'|'success'; description: string;
```
samuelli: Did auto formatter run?

src/server/controllers/webhook-events.ts
```diff
@@ -21,8 +26,76 @@ type PullRequestReviewHook = {
...
+  repository: RepositoryHook;
+  commit: {author: User;}
+};
+
 // Triggered when the status of a Git commit changes.
-export async function handleStatus(_hookBody: string) {
+export async function handleStatus(hookData: StatusHook) {
+  if (hookData.state !== 'error' && hookData.state !== 'failure') {
+    return;
```
samuelli: It might be useful from a debugging perspective if we returned different 2xx codes depending on what we do. eg. 202 on cases where we don't do anything with the webhook event.

(Don't need to address in this PR).

src/sw/sw.ts
```diff
@@ -18,4 +18,23 @@ const pushEventHandler = {
...
+};
+
+const clickEventHandler = {
+  handleEvent: (event: NotificationEvent) => {
+    event.notification.close();
+
+    // tslint:disable-next-line:no-any
+    const notification = ((event.notification as any) as CustomNotification);
+    if (notification.data && notification.data.url) {
+      event.waitUntil(self.clients.openWindow(notification.data.url));
```
samuelli: What does this do?

gauntface: I've broken this line up and commented it. Essentially it opens a window and ensures the SW isn't killed until the window has opened.

PR #148, opened 2018-02-12
Moving user tokens into firestore

gauntface:
This moves tokens into firestore rather than keeping them in memory on the server.

The original desire from everyone was to have tokens nested under a user, for example:

> users > ${username} > tokens > ${token} > ${tokenDetails}

This presents a problem when looking up information when we only have the token. Firestore only supports simply search of document fields on a collection (i.e. search for a user or search for token details).

I ended up using:

> tokens > ${token} > ${token Details}

This can be search for as `Find tokens where the username == 'gauntface'` allowing information look up by a token or a users login.

### DIFF
--- a/src/server/models/userModel.ts
+++ b/src/server/models/userModel.ts
@@ -2,31 +2,42 @@ import * as express from 'express';
 import gql from 'graphql-tag';
 
 import {ViewerLoginQuery} from '../../types/gql-types';
-import {github} from '../../utils/github';
 import {firestore} from '../../utils/firestore';
+import {github} from '../../utils/github';
+
+const TOKEN_COLLECTION_NAME = 'githubTokens';
 
 interface LoginDetails {
   username: string;
-  token: string;
+  githubToken: string;
   scopes: string[]|null;
 }
 
+/**
+ * The structure of the data base is:
+ *
+ * - githubTokens
+ *   - <GitHub Token>
+ *     - username
+ *     - githubToken
+ *     - scopes
+ */
 class UserModel {
   async getLoginDetails(login: string): Promise<LoginDetails|null> {
-    const tokensCollection = await firestore()
-        .collection('tokens');
+    const tokensCollection =
+        await firestore().collection(TOKEN_COLLECTION_NAME);
 
     const query = await tokensCollection.where('username', '==', login).get();
     if (query.empty) {
       return null;
     }
-    
+
     const loginDetails = query.docs[0].data();
     if (!loginDetails) {
       return null;
     }
-    
-    return (loginDetails as LoginDetails);
+
+    return loginDetails as LoginDetails;
   }
 
   async getLoginFromRequest(request: express.Request):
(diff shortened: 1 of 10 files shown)


src/server/models/userModel.ts
```diff
@@ -58,13 +65,17 @@ class UserModel {
...
+
     const details = {
       username: loginResult.data.viewer.login,
-      token,
       scopes,
+      token,
     };
-
-    this.cachedLogins[token] = details;
+
```
samuelli: nit: trailing whitespace - did formatter run?

src/server/models/userModel.ts
```diff
@@ -11,21 +12,21 @@ interface LoginDetails {
...
+      return null;
+    }
+    
+    const loginDetails = query.docs[0].data();
+    if (!loginDetails) {
+      return null;
     }
-    return null;
+    
+    return (loginDetails as LoginDetails);
```
aomarks: no need for parens

src/server/models/userModel.ts
```diff
@@ -11,21 +12,21 @@ interface LoginDetails {
 }
 
 class UserModel {
-  private cachedLogins: {[id: string]: LoginDetails};
+  async getLoginDetails(login: string): Promise<LoginDetails|null> {
+    const tokensCollection = await firestore()
```
aomarks: Unclear what kind of tokens these are. Can you add "github" to the naming in the db and vars here?

src/server/models/userModel.ts
```diff
@@ -11,21 +12,21 @@ interface LoginDetails {
...
-    this.cachedLogins = {};
-  }
-
-  async getLoginDetails(username: string) {
-    for (const token of Object.keys(this.cachedLogins)) {
-      const details = this.cachedLogins[token];
-      if (details.username === username) {
-        return details;
-      }
+    const query = await tokensCollection.where('username', '==', login).get();
```
aomarks: FTR every document field is automatically indexed, so this should perform well: https://firebase.google.com/docs/firestore/query-data/indexing

src/server/models/userModel.ts
```diff
@@ -11,21 +12,21 @@ interface LoginDetails {
 }
 
 class UserModel {
-  private cachedLogins: {[id: string]: LoginDetails};
+  async getLoginDetails(login: string): Promise<LoginDetails|null> {
```
aomarks: It was nice in your subscriptions PR that you had a block comment describing the database layout. Can you do the same here somewhere?

aomarks:
> This presents a problem when looking up information when we only have the token. Firestore only supports simply search of document fields on a collection (i.e. search for a user or search for token details).

What's the use case where we only have the token?

gauntface:
@samuelli and @aomarks should be good for review.

The two scenarios we need information are:
- Browser makes a request and the only information we know if the Github Token via the request cookie.
- Github WebHook that provides a Github username and we need to get a token related to it.

samuelli:
> Browser makes a request and the only information we know if the Github Token via the request cookie.

FWIW, this is a self imposed requirement. We could also store their GitHub login in the cookie alongside the token. However, we cannot trust the login provided in the cookie, so we would need to verify it against the tokens we have.

gauntface:
I think it's a sane pattern to follow for the reason you've pointed out and if the reason I've structure the data this way.

PR #153, opened 2018-02-13
Non module support

gauntface:
This PR changes how Typescript is built for the client.

Instead of using just the typescript compiler, this runs the client typescript files through rollup with a typescript plugin to handle transpilation of TS to ES2015.

- This will build all files in `src/client/`. Rollup is generally designed to have "entry" points which we can use in the future if we want, but for now, this seems like most flexible - just build everything as an entry point.
- Also fix the breaking change I made to the dashbaord.


gulp-tasks/build-client.js
```diff
@@ -1,11 +1,12 @@
 const path = require('path');
-const {buildTypescript} = require('./utils/build-typescript');
+const {buildBrowserTypescript} = require('./utils/build-browser-typescript');
 
 function buildClient() {
-  const destDir = path.join(global.__buildConfig.dest, 'client')
```
gauntface: Adding my own comment to gather test data.

samuelli:
LGTM, but deferring to @aomarks. PTAL.

gulp-tasks/utils/build-browser-typescript.js
```diff
@@ -0,0 +1,73 @@
...
+    input: scriptPath,
+    output: {
+      format: 'iife',
+      sourcemap: true,
+      name: path.basename(scriptPath),
+    },
+    plugins: [
+      typescriptPlugin({
+        ...tsconfig.compilerOptions,
+        tsconfig: false,
```
aomarks: This ignores options in the tsconfig -- that doesn't seem right?

https://github.com/rollup/rollup-plugin-typescript#usage

aomarks: Oh sorry ignore me, missed the line right above.

PR #159, opened 2018-02-14
Redirect users that need to login when navigating

gauntface:
- Moves client files into two directories:
   - client/public/
   - client/require-login/
- Dash-Server changed such that only `public/` and `/api/login` can be accessed without the id cookie.
- Enables pretty urls `/dash` `/settings` etc (and can still be accessed with .html)
- Moves `/dash.json` to `/apis/dash.json` just to be inline with other post API calls


src/client/public/scripts/dash.ts
```diff
@@ -181,9 +181,10 @@ function prTemplate(pr: api.PullRequest) {
 
 async function start() {
   const queryParams = new URLSearchParams(window.location.search);
+  // This allows your to see another users dashboard.
```
aomarks: s/your/you

src/server/apis/login.ts
```diff
@@ -29,19 +29,25 @@ function getRouter(): express.Router {
...
-            loginResponseBody['scope'].split(',') :
-            [];
-        await userModel.addNewUser(accessToken, userScopes);
+        try {
+          const accessToken = loginResponseBody['access_token'];
+          const userScopes = loginResponseBody['scope'] ?
+              loginResponseBody['scope'].split(',') :
+              [];
+              
+          await userModel.addNewUser(accessToken, userScopes);
```
aomarks: Can you scope the `try` to just this line?

PR #160, opened 2018-02-14
Adding TTL to web push to avoid old notifications forever trying

gauntface:
I opened up and older version of chrome and got a tonne of notifications.

This makes notifications get dropped if not delivered in 24 hours.

Rationale is that most of us will open github-health at least once a day, so if the notifications haven't been delivered in a day, probably not worth bombarding the user.

### DIFF
--- a/src/types/web-push.d.ts
+++ b/src/types/web-push.d.ts
@@ -6,5 +6,5 @@ declare module 'web-push' {
       ): void;
 
   export function sendNotification(
-      subscription: object, payload: string): Promise<void>;
+      subscription: object, payload: string, options: object): Promise<void>;
 }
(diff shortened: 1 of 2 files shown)


src/types/web-push.d.ts
```diff
@@ -6,5 +6,5 @@ declare module 'web-push' {
       ): void;
 
   export function sendNotification(
-      subscription: object, payload: string): Promise<void>;
+      subscription: object, payload: string, options: object): Promise<void>;
```
aomarks: can you define an inline type for options?

src/types/web-push.d.ts
```diff
@@ -1,3 +1,7 @@
+type WebPushOptions = {
+    TTL: number;
```
aomarks: it's optional right? `TTL?: number`

PR #161, opened 2018-02-15
Adding Fav Polling

gauntface:
This adds polling to the dashboard (updated every 30 minutes). This will change the favicon if there is a new actionable item.

### DIFF
--- a/src/client/public/scripts/dash/dash.ts
+++ b/src/client/public/scripts/dash/dash.ts
@@ -1,11 +1,10 @@
 import {html, render} from '../../../../../node_modules/lit-html/lit-html.js';
 import * as api from '../../../../types/api';
-import { PullRequest } from '../../../../types/api';
+import {PullRequest} from '../../../../types/api';
 
-// Poll every 30 Minutes
-const POLLING_INTERVAL = 30 * 60 * 1000;
+// Poll every 5 Minutes
+const POLLING_INTERVAL = 5 * 60 * 1000;
 
-let workLock = false;
 let lastResponse: api.DashResponse;
 
 type EventDisplay = {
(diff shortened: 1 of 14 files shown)


src/client/public/scripts/dash/dash.ts
```diff
@@ -1,5 +1,12 @@
-import {html, render} from '../../../../node_modules/lit-html/lit-html.js';
-import * as api from '../../../types/api';
+import {html, render} from '../../../../../node_modules/lit-html/lit-html.js';
+import * as api from '../../../../types/api';
+import { PullRequest } from '../../../../types/api';
```
samuelli: nit: formatter

src/client/public/scripts/dash/dash.ts
```diff
@@ -179,23 +199,113 @@ function prTemplate(pr: api.PullRequest) {
...
+    workLock = true;
+  
+    try {
+      await performPollAction();
+    } catch (err) {
+      console.log('Unable to perform poll update: ', err);
+    }
+  
+    workLock = false;
+  }, POLLING_INTERVAL);
```
samuelli: Using `setTimeout` might avoid the need for the workLock? Not entirely convinced the lock is needed.

src/client/public/scripts/dash/dash.ts
```diff
@@ -1,5 +1,12 @@
-import {html, render} from '../../../../node_modules/lit-html/lit-html.js';
-import * as api from '../../../types/api';
+import {html, render} from '../../../../../node_modules/lit-html/lit-html.js';
+import * as api from '../../../../types/api';
+import { PullRequest } from '../../../../types/api';
+
+// Poll every 30 Minutes
+const POLLING_INTERVAL = 30 * 60 * 1000;
```
samuelli: 30 seems really long. The dashboard needs to be pretty timely, so I would prefer if this was significantly shorter (in the order of minutes).

gauntface: 5 Minutes? I just don't want to basically kill a users quota.

samuelli: 5 is good for now. If this becomes problematic, we can look into reducing our query costs.

PR #173, opened 2018-02-16
Refactor typings for test contexts

samuelli:
Updates to pre-release version of Ava and follows new pattern for typing test contexts in Typescript https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md

### DIFF
--- a/src/test/cli/metrics/issue-count-test.ts
+++ b/src/test/cli/metrics/issue-count-test.ts
@@ -5,8 +5,7 @@ import {getIssueCounts} from '../../../cli/metrics/issue-counts';
 import {startTestReplayServer} from '../../../replay-server';
 import {initGithub} from '../../../utils/github';
 
-// tslint:disable-next-line:no-any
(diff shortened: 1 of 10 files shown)


src/test/cli/metrics/issue-count-test.ts
```diff
@@ -1,9 +1,13 @@
-import test from 'ava';
+import anyTest, {TestInterface} from 'ava';
+import {Server} from 'http';
 
 import {getIssueCounts} from '../../../cli/metrics/issue-counts';
 import {startTestReplayServer} from '../../../replay-server';
 import {initGithub} from '../../../utils/github';
 
+// tslint:disable-next-line:no-any
+const test: TestInterface<{server: Server}> = anyTest as any;
```
gauntface: can you add a comment on what `anyTest` is.

samuelli: I don't think it's anything actually. Its just the default export and I think this line just recasts it to have the right context types.

PR #178, opened 2018-02-17
Project health token

gauntface:
This stops us from stashing a GitHub token in a cookie and instead generates a custom user token that can be used to reference a users GitHub username and token.

### DIFF
--- a/src/server/apis/dash-data.ts
+++ b/src/server/apis/dash-data.ts
@@ -4,7 +4,7 @@ import gql from 'graphql-tag';
 import * as api from '../../types/api';
 import {mentionedFieldsFragment, prFieldsFragment, PullRequestReviewState, reviewFieldsFragment, ViewerPullRequestsQuery} from '../../types/gql-types';
 import {github} from '../../utils/github';
-import {userModel} from '../models/userModel';
+import {ID_COOKIE_NAME, userModel} from '../models/userModel';
 
 export class DashData {
   getHandler() {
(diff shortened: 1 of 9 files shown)


src/test/server/models/userModel-test.ts
```diff
@@ -0,0 +1,156 @@
...
+          .doc(TEST_USER_TOKEN)
+          .set({
+            username: TEST_USERNAME,
+          });
+
+      const result = await userModel.getLoginFromToken(TEST_USER_TOKEN);
+      t.deepEqual(result, null);
+    });
+
+test.serial('should return logindetails for token', async (t) => {
```
samuelli: login details

samuelli: I'm not sure this is a great test. It's very tightly coupled with the implementation instead of testing the high level API. I would prefer if this final test `getLoginFromToken()` was moved to the end of `'should return a new user token'`.

src/server/apis/dash-data.ts
```diff
@@ -4,15 +4,16 @@ import gql from 'graphql-tag';
...
+import {ID_COOKIE_NAME, userModel} from '../models/userModel';
 
 export class DashData {
   getHandler() {
     return this.handler.bind(this);
   }
 
   private async handler(req: express.Request, res: express.Response) {
-    const loginDetails = await userModel.getLoginFromRequest(req);
+    const loginDetails =
```
samuelli: It might make sense to add a new API function to `userModel` `getLoginFromRequest()` which just factors out reading from the cookie and calling `getLoginFromToken()`. WDYT?

PR #179, opened 2018-02-20
Split up dash data test

samuelli:


src/replay-server.ts
```diff
@@ -23,13 +23,14 @@ let recordingLogged = false;
...
-    Promise<{server: http.Server, url: string}> {
+export async function startTestReplayServer(
+    t: ava.ExecutionContext,
+    title?: string): Promise<{server: http.Server, url: string}> {
   const record = process.env.RECORD === 'true';
 
   // Kind of weirdly, t.title will include the name of the current function
   // plus " for " as a prefix. Get the original name instead.
-  const testTitle = t.title.replace(/^[\w\s]+ hook for /, '');
+  const testTitle = title ? title : t.title.replace(/^[\w\s]+ hook for /, '');
```
aomarks: This line is now a no-op, since you only run the replace when there is nothing to replace.

src/test/server/apis/dash-data-test.ts
```diff
@@ -1,259 +1,299 @@
...
-            createdAt: 1517253614000,
-            reviewState: PullRequestReviewState.APPROVED,
-          }],
-        }],
-      },
-      // Outgoing PR, has 1 commented review.
+test('dashdata: outgoing PRs are sorted', (t) => {
+  const data = t.context.dashData;
+  let lastCreatedAt = data.outgoingPrs[0].createdAt;
+  data.outgoingPrs.forEach((pr) => {
```
aomarks: why not use a `for of`?

src/replay-server.ts
aomarks: Oh, nevermind, misread. Now I'm just a little confused about what why there's both title and t.title.

PR #184, opened 2018-02-21
Sort incoming pull requests

### LINKED ISSUE #100: Sorting of incoming PRs
eg. sort by status, the recency etc

samuelli:
This change sorts the list of incoming pull requests that are displayed. They are ordered first by whether or not the status is actionable. Next they are sorted by the latest event time, which excludes your review event if any.

This allows new activity to bubble to the top, while maintaining an actionable view.

Closes #100.


src/server/apis/dash-data.ts
```diff
@@ -231,11 +231,67 @@ export class DashData {
...
+    }
+    return latest;
+  }
+
+  // Sort descending by latest event time.
+  const timeSorted =
+      prs.sort((a, b) => getLatestEventTime(b) - getLatestEventTime(a));
+  const result = [];
+
+  const notActionable = [
```
gauntface: Out of the scope for this PR, but does it make sense to add an `actionable` field to this data? I just worry that any changes to this list will need changes in a few different places.

src/server/apis/dash-data.ts
```diff
@@ -231,11 +231,67 @@ export class DashData {
...
+    'UnknownStatus',
+    'NoActionRequired',
+    'NewActivity',
+    'StatusChecksPending'
+  ];
+  // Add actionable PRs.
+  for (const pr of timeSorted) {
```
gauntface: You could maintain to arrays (actionable / not-actionable) and concat at the end - avoiding the double loop.

src/server/apis/dash-data.ts
samuelli: Yeah maybe. Originally I wanted to keep it out of the server, but now its in the server.

src/server/apis/dash-data.ts
samuelli: done.

Issue #183, opened 2018-02-21
Login error


samuelli:
![image](https://user-images.githubusercontent.com/787668/36499153-3bca2228-16f5-11e8-859c-4860f5b59b25.png)

![image](https://user-images.githubusercontent.com/787668/36499169-4c02f2fa-16f5-11e8-84be-b7b0b3a500eb.png)

PR #185, opened 2018-02-21
Basic profile

### LINKED ISSUE #169: Implement profile header
image

For now, settings button on the right that points to the settings page.

In the future, maybe it is a profile switcher. TBD.

gauntface:
Fixes #169

This is super basic place for the profile.

I also hit an issue where the login details were available, but I had changed the details requested and needed a way to "validate" the user login details, so I've added a simple set of checks to the user model and this might address #183 (Login Error)


src/client/public/styles/components/profile.css
```diff
@@ -0,0 +1,43 @@
...
+  border-radius: 4px;
+  padding: var(--padding);
+}
+
+.profile-avatar {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+
+  width: calc(var(--padding) * 4);
```
samuelli: Can we make it a bit bigger?

src/client/public/styles/variables/colors.css
```diff
@@ -1,6 +1,10 @@
 :root {
+  --light-grey: #aaaaaa;
```
samuelli: hsl(0, 0%, 80%)

PR #186, opened 2018-02-22
Api strict

### LINKED ISSUE #183: Login error

gauntface:
After getting stuck in some nasty error logic for logins I wanted to clean up the response.

This essentially the start of a broader refactor where all JSON API responses return a valid JSON response of either `data` or `error`. The data can be any type, error should be a code and message.

This only moves the login over to this approach, but if it looks ok, I'll refactor other parts of the code. 

While #183 was fixed by pushing latest changes to prod, this will vastly improve the debugging and in my eyes fixes #183.


src/client/public/oauth.html
```diff
@@ -1,50 +1,3 @@
...
-          `https://github-health.appspot.com/oauth.html?redirect-origin=${REDIRECT_ORIGIN}&final-redirect=${queryParams.get('final-redirect') || ''}`
-        ),
-      };
-      const searchValues = Object.keys(properties).map((keyName) => {
-        return `${keyName}=${properties[keyName]}`;
-      });
-      window.location = `https://github.com/login/oauth/authorize?${searchValues.join('&')}`;
-    }
-  })();
 </script>
```
samuelli: File empty now?

PR #191, opened 2018-02-22
Replace grid layout with flexbox

samuelli:
This change removes the existing grid layout and replaces it primarily with flexbox. This allows us to create logical containers for layout around pull requests and timeline events for pull requests. This change also removes lines from the timeline bullets and instead replaces it with a single line. As such, events can now flow onto multiple lines without breaking the timeline bullet UI.

image

### DIFF
--- a/src/client/public/styles/dash.css
+++ b/src/client/public/styles/dash.css
@@ -16,8 +16,6 @@ a {
 
 .pr {
   position: relative;
-  --avatar-size: 40px;
-  --time-width: 100px;
 }
 
 .pr::before {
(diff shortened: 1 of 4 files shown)


src/client/public/styles/dash.css
```diff
@@ -7,38 +7,55 @@ a {
...
-  grid-gap: 0px var(--padding);
-
   font-size: 14px;
 }
 
-.pr-author {
-  grid-column: 1 / 1;
+.pr {
+  position: relative;
+  --avatar-size: 40px;
```
gauntface: I would put this into the `dimens.css` file. Rationale being that if there are common patterns for the UI for follow - it's clearer if the values are in one spot - it also encourages re-use, which may help with future refactors.

samuelli: Done.

PR #192, opened 2018-02-22
Ordering settings and updating toggles to match designs

### LINKED ISSUE #182: Sort orgs in settings, and update text
- Text should change when toggling between two states
- Orgs should be sorted by can toggle, vs can't toggle.

gauntface:
Fixes #182


src/client/public/scripts/settings/org-webhook-list.ts
```diff
@@ -39,14 +39,21 @@ function hookTemplate(org: OrgWebHookState) {
       checkboxElement.checked = !checkboxElement.checked;
     }
 
-    checkboxElement.removeAttribute('disabled');
+    updateUI();
+    // checkboxElement.removeAttribute('disabled');
```
samuelli: Remove

PR #193, opened 2018-02-22
Start of flashing

### LINKED ISSUE #165: Flash of color when focusing page with new actions

gauntface:
Fixes #165
Sam's styles applied - I extended the delay by an extra 200ms just to make it easier to see before the transition out.

I had to fiddle with the transitions in CSS to only apply when fading *out*. Previously it was fading in  and the results where extremely unreliable when navigating to the page.

### DIFF
--- a/src/client/public/scripts/dash/dash.ts
+++ b/src/client/public/scripts/dash/dash.ts
@@ -368,19 +368,13 @@ async function start() {
     // frame to ensure the class is up to date, then apply the flash.
     const elements = document.querySelectorAll('.is-newly-actionable');
     for (let i = 0; i < elements.length; i++) {
-      const element = elements.item(i);
+      const element = elements.item(i) as HTMLElement;
       element.classList.remove('actionable-flash');
+      // tslint:disable-next-line:no-unused-expression
+      element.offsetTop;  // Force a style recalc
+      element.classList.remove('is-newly-actionable');
+      element.classList.add('actionable-flash');
     }
-
-    window.requestAnimationFrame(() => {
-      const elements = document.querySelectorAll('.is-newly-actionable');
-
-      for (let i = 0; i < elements.length; i++) {
-        const element = elements.item(i);
-        element.classList.remove('is-newly-actionable');
-        element.classList.add('actionable-flash');
-      }
-    });
   });
 
   performPolling();
--- a/src/client/public/scripts/dash/dash.ts
+++ b/src/client/public/scripts/dash/dash.ts
@@ -20,21 +20,25 @@ type StatusDisplay = {
 
 const DEFAULT_AVATAR = '/images/default-avatar.svg';
 
-const dashTmpl = (data: api.DashResponse) => {
+const dashTmpl = (data: api.DashResponse, newlyActionablePRs: string[]) => {
   const imageUrl = data.user.avatarUrl ? data.user.avatarUrl : DEFAULT_AVATAR;
   const buttonTemplates = [];
+
   if (data.user.isCurrentUser) {
     buttonTemplates.push(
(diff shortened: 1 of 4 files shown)


src/client/public/scripts/dash/dash.ts
```diff
@@ -324,15 +335,29 @@ function performPolling() {
...
+    // When an element is marked as 'is-newly-actionable' in the template
+    // it will also have the 'u-no-tranisitions' class applied. This means
+    // the browser will render the background with no transition.
+    // We then remove the u-no-transitions class and update the rendered
+    // template to transition out the flash
+    const elements = document.querySelectorAll('.is-newly-actionable');
+    for (let i = 0; i < elements.length; i++) {
+      const element = elements.item(i);
+      element.classList.remove('u-no-tranisitions');
+    }
```
samuelli: The more typical approach to this problem is to use CSS animations.

```
@keyframes flash {
  from {
     background-color: var(--actionable-highlight-color);
  }
  to {
    background-color: inherit;
  }
}
```

To ensure you can animate & restart the same animation, cause as style recalc:
```
element.classList.remove('flashClass'); // Ensure its not already animated
element.style.offsetTop; // Force a style recalc
element.classList.add('flashClass');
```

It's not too dissimilar from the approach here, but it'll reduce a little bit of the complexity: only one CSS class needed, only one point of logic.

gauntface:
@samuelli that is much nicer - only tweak to your suggestion was to use request animation frame instead of forcing a style recalculation and I added a delay of the color animation. (I also had to use a color instead of inherit as inherit displays the line under the bullets for some reason).

src/client/public/styles/dash.css
```diff
@@ -27,7 +27,8 @@ a {
   top: calc(var(--padding) / 3);
   bottom: calc(var(--padding) / 3);
   left: calc(var(--time-width) + var(--padding) + var(--avatar-size) / 2 - 1px);
-  z-index: -1;
+  /* 1px to ensure the line is *below* the avatar */
+  margin-top: 1px;
```
samuelli: Just `+ 1` to the `top` calc above.

src/client/public/styles/dash.css
```diff
@@ -131,8 +133,25 @@ a {
...
+  background-color: var(--page-bg-color);
+}
+
+@keyframes flashKeyframe {
+  0% {
+    background-color: var(--actionable-highlight-color);
+  }
+  25% {
+    background-color: var(--actionable-highlight-color);
+  }
```
samuelli: Remove this keyframe.

src/client/public/styles/dash.css
```diff
@@ -131,8 +133,25 @@ a {
 }
 
 /* Hide the timeline line after the last event  */
-.pr > .pr-header:last-child, .pr > .pr-event:last-child > .pr-event__bullet svg {
-  background: white;
+.pr, .pr > .pr-event:last-child > .pr-event__bullet svg {
+  background-color: var(--page-bg-color);
+}
+
+@keyframes flashKeyframe {
```
samuelli: It's a (keyframed) animation, name is slightly incorrect.

src/client/public/styles/dash.css
```diff
@@ -131,8 +133,25 @@ a {
...
+    background-color: var(--actionable-highlight-color);
+  }
+  to {
+    background-color: var(--page-bg-color);
+  }
+}
+
+.pr.actionable-flash, .pr.actionable-flash > .pr-event:last-child > .pr-event__bullet svg {
+  animation-duration: 700ms;
+  animation-name: flashKeyframe;
```
samuelli: Just add a `animation-delay` here. You may also need `animation-fill-mode: backwards` as well.

PR #200, opened 2018-02-23
Tidy up of push sw

### LINKED ISSUE #198: Support Refreshing Dashboard on Web Push

gauntface:
Fixes #198

I took this opportunity to clean up the SW in general.

Now when a notification comes in a few things occur that didn't before:

- Default notifications are used *if* the received push is bad. This is a last resort and shouldn't ever get used.
- No notification is shown if a window is currently visible.
- Messages are sent to every open window informing them that a push was received. Dash windows will listen for this and update upon receiving this.

On notification clicks:

- We will now look for a currently open window and focus it if the URLs match, otherwise, we'll open a new window (NOTE: This will only work for urls on the same origin - i.e. if we ever link to github.com, this will always open a new window.

### DIFF
--- a/src/client/public/scripts/dash/dash.ts
+++ b/src/client/public/scripts/dash/dash.ts
@@ -1,8 +1,6 @@
 import {html, render} from '../../../../../node_modules/lit-html/lit-html.js';
 import * as api from '../../../../types/api';
 
-import {JSONAPIResponse, LastKnownResponse, PullRequest, SWClientMessage} from '../../../../types/api';
-
 // Poll every 5 Minutes
 const LONG_POLL_INTERVAL = 5 * 60 * 1000;
 const SHORT_POLL_INTERVAL = 15 * 1000;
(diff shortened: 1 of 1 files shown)


src/client/public/scripts/dash/dash.ts
```diff
@@ -1,6 +1,7 @@
 import {html, render} from '../../../../../node_modules/lit-html/lit-html.js';
 import * as api from '../../../../types/api';
-import {JSONAPIResponse, LastKnownResponse, PullRequest} from '../../../../types/api';
+
+import {JSONAPIResponse, LastKnownResponse, PullRequest, SWClientMessage} from '../../../../types/api';
```
samuelli: This has 2 imports form `api` here. Do you mind cleaning that up here? Either the second style where we specify exactly what we want from the API, or the first where we import everything and use `api.*` everywhere.

src/client/public/scripts/dash/dash.ts
```diff
@@ -426,9 +431,20 @@ async function start() {
...
-  startPolling(queryParams.get('login'));
+  if (navigator.serviceWorker) {
+    navigator.serviceWorker.addEventListener('message', (event) => {
+      if (!event.data || !event.data.action) {
+        return;
+      }
+
+      const swMessage = event.data;
+      if (swMessage.action === 'push-received') {
+        performLongPoll(userLogin);
```
samuelli: (optional) `performLongPoll` is the fn which performs the dashboard update right? Don't really like the name, `updateDashboard()` would make more intuitive sense.

PR #202, opened 2018-02-24
Integration tests by running pixel tests using puppeteer

### LINKED ISSUE #104: Pixel test for client
Use headless, use replay data, never hit the network

samuelli:
This change introduces new integration tests for the client and server. The pixel test runner exposes a simple API which expects a Puppeteer page that is ready for screenshots.
 - If no golden is found, the runner tells you exactly what command to run.
 - If the output doesn't match the golden, the runner will generate a side by side comparison of the images (actual, expected, diff) and outputs a) the command to view this side by side and b) the command to rebaseline the test if necessary.

Closes #104.


src/pixel-tester.ts
```diff
@@ -0,0 +1,107 @@
...
+    console.info(`💾 Generated golden at ${expectedPath}`);
+  }
+}
+
+/**
+ * Promisified version of writing a jimp to a file.
+ */
+function writeJimp(jimp: jimp, path: string): Promise<void> {
+  return new Promise((resolve) => {
+    jimp.write(path, () => resolve());
```
aomarks: `jimp.write(Path(), resolve)`

gauntface: nit: I'd also change the variable name from `path` to something else - just to avoid confusion with the node `path` module.

src/pixel-tester.ts
```diff
@@ -0,0 +1,107 @@
+import * as ava from 'ava';
+import * as fs from 'fs-extra';
+import * as jimp from 'jimp';
+import * as path from 'path';
+import {PNG} from 'pngjs';
+import * as puppeteer from 'puppeteer';
+
+// tslint:disable-next-line:no-require-imports
+import pixelmatch = require('pixelmatch');
```
gauntface: maybe add a comment on why this is require instead an import

src/pixel-tester.ts
gauntface: would `utils.promisify` work here? https://nodejs.org/dist/latest-v8.x/docs/api/util.html

samuelli: Tried `util.promisify` but TS isn't happy. Changed variable name to `imgPath` instead.

src/pixel-tester.ts
samuelli: Done.

PR #210, opened 2018-02-28
Adding activity notification to dashboard

### LINKED ISSUE #208: New activity notification
After the dashboard refreshes, we check the last time the dashboard was checked. If its more than an hour ago, we show the following notification:
image

We continue to do this on every new activity we have. The notification does *not* require interaction. Because its done after polling, this will only ever be shown if the dashboard is open.

gauntface:
Fixes #208

Using a new timestamp as this will be the users machines timestamp and NOT the server timestamp.

Updates once an hour.

### DIFF
--- a/src/client/public/scripts/dash/dash.ts
+++ b/src/client/public/scripts/dash/dash.ts
@@ -1,9 +1,13 @@
 import {html, render} from '../../../../../node_modules/lit-html/lit-html.js';
 import * as api from '../../../../types/api';
 
-// Poll every 5 Minutes
+// Poll every 5 minutes
 const LONG_POLL_INTERVAL = 5 * 60 * 1000;
+// Poll every 15 seconds
 const SHORT_POLL_INTERVAL = 15 * 1000;
+// Wait 1 hour before showing local notification if updates have occured.
+const ACTIVITY_UPDATE_DURATION = 1 * 60 * 60 * 1000;
+
 const NO_OUTGOING_PRS_MESSAGE =
     'You have no outgoing pull requests. When you open new pull requests, they\'ll appear here';
 const NO_INCOMING_PRS_MESSAGE =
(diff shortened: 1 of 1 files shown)


src/client/public/scripts/dash/dash.ts
```diff
@@ -383,6 +392,58 @@ async function performShortPollAction(userLogin: string|null) {
...
 
+async function hasPushEnabled() {
+  if ('permissions' in navigator) {
+    // tslint:disable-next-line:no-any
+    const permissionsAPI = (navigator as any)['permissions'];
+    const result = await permissionsAPI.query({
+      name: 'push',
+      userVisibleOnly: true,
+    });
+    if (result.state === 'granted') {
```
samuelli: `return result.state === 'granted'`

src/client/public/scripts/dash/dash.ts
```diff
@@ -383,6 +392,58 @@ async function performShortPollAction(userLogin: string|null) {
...
+    return;
+  }
+
+  // Reset the timestamp to wait a new hour before showing notification
+  lastActivityUpdateTimestamp = Date.now();
+  if (!await hasPushEnabled()) {
+    return;
+  }
+
+  const newlyActionablePRs = hasNewActions(lastPolledData, lastViewedData);
```
samuelli: This will change in the refactor and one of us will have to rebase.

src/client/public/scripts/dash/dash.ts
```diff
@@ -383,6 +392,58 @@ async function performShortPollAction(userLogin: string|null) {
...
+  if (newlyActionablePRs.length < 0) {
+    return;
+  }
+
+
```
samuelli: nit: probably don't need these double new lines here and below.

src/client/public/scripts/dash/dash.ts
```diff
@@ -414,6 +475,7 @@ async function start() {
     changeFavIcon(false);
 
     lastViewedData = lastPolledData;
+    lastActivityUpdateTimestamp = Date.now();
```
samuelli: So is the timestamp from the server ignored now? If so, should we just remove it from there?

gauntface: The two diff timestamps make sense, but with your changes, it might make sense to just remove server timestamp and use local time instead.

gauntface: Circling back on this.

The timestamps from the server are still needed.

We short poll our server and ask "what was the last known update (via github hook) that we heard about (using the servers timestamp)" and then we compare that to the last timestamp that we hard for a full dash load".

The local activity is, when has it been longer than an hour with the dashboard open. This is seperate form the server time.

PR #224, opened 2018-03-02
Adding all information needed to offer a merge UI in the client

gauntface:
With the combination of (mergeable, repoDetails.allow_*merge and status, the API should be able to render the auto merge, which I'll do in a follow up PR).

This runs through each PullRequest and creates an OutgoingPullRequest in parallel as we discussed.


src/server/apis/dash-data.ts
```diff
@@ -71,59 +72,78 @@ export async function fetchOutgoingData(
...
-  const outgoingPrs = [];
+  const outgoingPrs: api.OutgoingPullRequest[] = [];
   if (viewerPrsResult.data.user) {
-    for (const pr of viewerPrsResult.data.user.pullRequests.nodes || []) {
-      if (!pr) {
-        continue;
-      }
+    const requestPrs = viewerPrsResult.data.user.pullRequests.nodes || [];
+    const outgoingPrPromises =
+        requestPrs.map(async(pr): Promise<api.OutgoingPullRequest|null> => {
```
samuelli: Are we sure this runs in parallel?

gauntface: I believe it does yes. Is there a reason you think it's not?

PR #229, opened 2018-03-05
Move away from rollup typescript

gauntface:
This moves away from using rollup to convert typescript to JS and instead uses gulp-typescript (which seems to be more inline with tsc). Once the build has run, rollup is used to convert es2015 imports to a single bundle.

I've then moved our html pages to load up both module and non-module versions and the module sourcemaps seem to work with this (at least chrome is loading the .ts file as a source.


gulp-tasks/utils/build-browser-typescript.js
```diff
@@ -24,50 +24,45 @@ const processScript = (scriptPath, relativePath, tsConfigPath, destDir) => {
...
-             // in the plugin options (hence the `...tsconfig.compilerOptions`)
-             // Then `tsconfig: false` disables any tsconfig.json file reading
-             typescriptPlugin({
-               ...tsconfig.compilerOptions,
-               tsconfig: false,
-               typescript: require('typescript'),
-             }),
-             uglifyPlugin({}, esMinify),
+             sourcemapPlugin(),
+             // uglifyPlugin({}, esMinify),
```
samuelli: Should this be back?

gauntface: For now I'd say no - just to see if the sourcemap is even remotely useful.

Issue #166, opened 2018-02-15
Dashboard doesn't show all issues


samuelli:
Current idea: show dashboard as is. If we know there are more issues, then we can show "View all" buttons on each section which take you to separate pages. These just have infinite lists which show all the data and acts similar to the dashboard.

PR #242, opened 2018-03-07
Add view all outgoing page

samuelli:
This new page is served from `/outgoing` and shows all the outgoing pull requests by performing async requests against the API.

Part of #166


src/client/public/scripts/dash/outgoing.ts
```diff
@@ -0,0 +1,35 @@
...
+    // Set new cursor information.
+    const cursorParam: string = data ? `cursor=${data.cursor}` : '';
+    const response = await fetch(
+        `/api/dash/outgoing?${[loginParam, cursorParam].join('&')}`,
+        {credentials: 'include'});
+    data = await response.json() as api.OutgoingDashResponse;
+
+    // Render each PR.
+    for (const pr of data.prs) {
+      yield prTemplate(pr, []);
```
gauntface: maybe the newly actionable PR array should be made optional.

src/client/public/scripts/dash/outgoing.ts
```diff
@@ -0,0 +1,35 @@
+import {asyncAppend} from '../../../../../node_modules/lit-html/lib/async-append.js';
+import {html, render} from '../../../../../node_modules/lit-html/lit-html.js';
+import * as api from '../../../../types/api';
+
+import {prTemplate} from './prs';
+
+async function* getNextElement(userLogin: string|null) {
```
gauntface: Might be worth adding a comment how this works with lit. I'm not familiar with generators and the use of `function*` and `yield` is new to me.

PR #244, opened 2018-03-07
Remove aomarks from codeowners.

aomarks:
Totally happy to do reviews when it helps, but right now I'm getting assigned to everything that goes through even when I don't feel like I am needed.


samuelli:
Hmm, I wonder if me as an owner, requesting a review from @aomarks is sufficient for GitHub to be happy. That's the Chromium model.

PR #248, opened 2018-03-07
Moving to tsc for all building and bundled builds second gulp step

gauntface:
- tsc used for builds
- tsc output commonjs for server, test etc and outputs es2015 modules for sw and client
    - .js files in the 1:1 mapping of src > build (i.e. src/sw/sw.ts -> build/sw/sw.js) will be es2015 modules
    - the <path>/bundled/**/*.js files will be bundled versions of the es2015 modules.
- Removed gql-types from api.ts and added a comment.


src/client/public/oauth.html
```diff
@@ -1,2 +1,2 @@
-<script nomodule src="/scripts/oauth/oauth.js" crossorigin></script>
-<script type="module" src="/scripts/oauth/oauth.mjs" crossorigin></script>
+<script nomodule src="/bundled/scripts/oauth/oauth.js" crossorigin></script>
+<script type="module" src="/scripts/oauth/oauth.js" crossorigin></script>
```
samuelli: This is also nicer I think - DevTools doesn't recognize `.mjs` and won't link to the source properly.

src/server/dash-server.ts
```diff
@@ -14,42 +15,74 @@ import {getRouter as getUpdatesRouter} from './apis/updates';
...
-    app.use(express.static(
-        path.join(__dirname, '..', 'client', 'public'),
-        {extensions: ['html', 'mjs']}));
-    app.use(express.static(path.join(__dirname, '..', 'sw')));
-    app.use('/api/login/', bodyParser.text(), getLoginRouter());
-    app.use('/api/webhook/', bodyParser.json(), getGitHubHookRouter());
-
-    // Require Login for all endpoints used after this middleware
+    app.use(bodyParser.text());
+    app.use(bodyParser.json());
```
samuelli: These don't fail if they're not JSON?

gauntface: No - it requires the content type header to be set correctly.

PR #273, opened 2018-03-08
Auto merge UI

gauntface:
This adds auto-merge UI to the dashboard. This allows selecting of merge type and will be up-to-date if refreshed.

Will add tests in a follow PR to ensure this is WAI and stays that way.


src/client/public/scripts/dash/auto-merge-events.ts
```diff
@@ -0,0 +1,165 @@
...
+  );
+}
+
+export function getAutoMergeOptions(pr: api.PullRequest): TemplateResult[] {
+  if (!('mergeable' in pr)) {
+    // This is not an outgoing PR
+    return [];
+  }
+
+  const outgoingPr = (pr as api.OutgoingPullRequest);
```
samuelli: Not a fan of this. We already knew the type of this, so it doesn't seem great to then merge and split it out again. Don't have a good idea of

PR #283, opened 2018-03-13
Making status take into account the PR states

gauntface:
This enables auto-merge UI with new repo that includes status for each PR and sets the correct status type for outgoing PR's


gauntface:
@samuelli I'm really struggling to make the pixel tests pass one after the other (even locally).

My hunch is that the test server is using the same name for the tests but the results are no difference because of the paging.

src/server/apis/dash-data/fetch-outgoing-data.ts
```diff
@@ -153,9 +153,27 @@ async function getAllPRInfo(
         });
       }
 
+      let latestCommit: commitFieldsFragment|null = null;
```
samuelli: Move this line down to where its used. L165.

src/server/apis/dash-data/fetch-outgoing-data.ts
```diff
@@ -260,7 +280,17 @@ function getStatus(
...
+        const state = latestCommit.status.state;
+        if (state === 'PENDING') {
+          outgoingStatus = {type: 'StatusChecksPending'};
+        } else if (state === 'ERROR' || state === 'FAILURE') {
+          outgoingStatus = {type: 'StatusChecksFailed'};
+        }
+      }
+      if (outgoingStatus.type === 'UnknownStatus') {
+        outgoingStatus = {type: 'PendingMerge'};
+      }
```
samuelli: What about when status checks are passing?

gauntface: the two remaining states are no states and success states, both should result in PendingMerge for the dashboard state.

samuelli:
> @samuelli I'm really struggling to make the pixel tests pass one after the other (even locally).
> 
> My hunch is that the test server is using the same name for the tests but the results are no difference because of the paging.

Just leave these then.

PR #312, opened 2018-03-20
Changing PR model to store with owner,repo and pr number

gauntface:
This moves from the Graphql PR ID to the owner~repo~number instead for firebase docs.


src/server/models/pullRequestsModel.ts
```diff
@@ -6,22 +6,77 @@ const PR_COLLECTION_NAME = 'pull-requests';
...
+      throw new Error('No repo.');
+    }
+
+    if (!num) {
+      throw new Error('No num.');
+    }
+    return `${owner}~${repo}~${num}`;
+  }
+
+  async openPullRequest(owner: string, repo: string, num: number) {
```
samuelli: `pullRequestOpened`?

src/server/models/pullRequestsModel.ts
```diff
@@ -6,22 +6,77 @@ const PR_COLLECTION_NAME = 'pull-requests';
...
+      throw new Error('No owner.');
+    }
+
+    if (!repo) {
+      throw new Error('No repo.');
+    }
+
+    if (!num) {
+      throw new Error('No num.');
+    }
```
samuelli: Num can't be null or empty? This would rule out 0 which may be ok, but probably unnecessary.

PR #329, opened 2018-03-21
Using analytics client ID in service worker events

gauntface:


src/client/public/scripts/models/client-id.ts
```diff
@@ -0,0 +1,73 @@
...
+               this.db = (event.target as any).result as IDBDatabase;
+               resolve(this.db);
+             };
+           }) as Promise<IDBDatabase>;
+  }
+
+  // tslint:disable-next-line:no-any
+  private async updateDB(event: any) {
+    const db = event.target.result;
+    if (event.oldVersion < 2) {
```
samuelli: What does this mean?

gauntface: If the oldVersion was < 2 (i.e. it's new), perform these steps and with future versions will have `< XYZ` and perform further upgrades.

Basically this sets up support for future changes.

samuelli: I see.

PR #346, opened 2018-03-22
Prevents handling of status events when the commit is a different SHA

### LINKED ISSUE #341: Notification on builds which aren't for the latest commit
## The Problem
 - Create a pull request which breaks the build
 - Add a commit that fixes the build
 - You'll receive notifications for the first commits builds even though a new commit has been pushed.

## Expected Behavior
You only receive notifications for builds that are the latest commit.

gauntface:
Fixes #341

The tests have been a growing source of pain for me so I ended up re-writing the tests to get coverage for the status hook up and to have more descriptive and easier to reason with tests.


src/test/server/controllers/webhook-events/status-test.ts
```diff
@@ -25,17 +26,95 @@ const TEST_SECRETS = {
...
+  },
+  commit: {
+    author: {
+      login: 'status-commit-author',
+    },
+  },
+};
+
+function newFakePRDetails(extraDetails: {}) {
+  const clone = Object.assign({}, PR_DETAILS);
```
samuelli: `return Object.assign({}, PR_DETAILS, extraDetails);`. Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

PR #362, opened 2018-03-27
Adding a training notification

### LINKED ISSUE #360: Training notification when notifications enabled
"You'll get a notification for important events"

gauntface:
Fixes #360

Note: This will appear every time push is enabled (i.e. toggling off and on will show the notification). This shouldn't be a problem as it's an unlikely user interaction, but it's worth calling out.

### DIFF
--- a/src/client/public/scripts/settings/push-component.ts
+++ b/src/client/public/scripts/settings/push-component.ts
@@ -120,7 +120,7 @@ class PushComponent {
         'You\'ll get notifications for important events', {
           icon: '/images/notification-images/icon-192x192.png',
           badge: '/images/notification-images/badge-128x128.png',
-          requiresInterfaction: true,
+          requiresInteraction: true,
           tag: 'training-notification',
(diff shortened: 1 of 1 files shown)


samuelli:
> Note: This will appear every time push is enabled (i.e. toggling off and on will show the notification). This shouldn't be a problem as it's an unlikely user interaction, but it's worth calling out.

I think this is a good/expected thing. The user wants confirmation that *something* happened.

src/client/public/scripts/settings/push-component.ts
```diff
@@ -116,6 +116,14 @@ class PushComponent {
     });
 
     await this.update();
+    await registration.showNotification(
+        'You\'ll get notifications for important events', {
+          icon: '/images/notification-images/icon-192x192.png',
+          badge: '/images/notification-images/badge-128x128.png',
+          requiresInterfaction: true,
```
samuelli: interaction, but probably shouldn't require interaction?

PR #372, opened 2018-03-28
Add a legend to each section which acts as a filter

samuelli:


src/client/public/scripts/dash/dash.ts
```diff
@@ -151,6 +169,57 @@ function onMessage(event: ServiceWorkerMessageEvent|MessageEvent) {
...
+
+  /**
+   * Finds and attaches filter event listener.
+   */
+  function attachFilterListener(id: FilterId) {
+    const element = document.getElementById(id);
+    if (!element) {
+      throw Error('Could not attach filter event listener');
+    }
+    element.addEventListener('legend-change', filterChanged.bind(null, id));
```
gauntface: You could dispatch an event on the window and use a data field to determine behavior - It's how the automerge behavior is currently working.

PR #378, opened 2018-03-30
Status success notification

gauntface:
The current "Ready to merge" notifications are misleading, so I've altered the logic to not send a notification if there are no approvals and to soften the language to be more inline with the current update.

### DIFF
--- a/src/server/utils/get-pr-from-commit.ts
+++ b/src/server/utils/get-pr-from-commit.ts
@@ -3,6 +3,11 @@ import gql from 'graphql-tag';
 import {CommitToPRQuery} from '../../types/gql-types';
 import {github} from '../../utils/github';
 
+export interface ReviewDetails {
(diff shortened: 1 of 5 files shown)


src/server/utils/get-pr-from-commit.ts
```diff
@@ -72,6 +79,25 @@ export async function getPRDetailsFromCommit(
...
+        if (review.state === 'APPROVED') {
+          approvedReviewCount++;
+        }
+
+        reviews.push({
+          state: review.state,
+          author: review.author.login,
+        });
+      }
+    }
```
samuelli: This doesn't use the same logic that we use on the main dashboard. This could lead to subtle differences. For example, some edge cases which this currently doesn't consider:
 - same reviewer approves multiple times. This would show up as n approved reviews, when really there was only 1.
 - reviewer changes their review - GitHub allows you to approve a change and then request changes.

The main dashboard deals with these cases by a) finding the _current_ review state for a reviewer and b) only looking at one review per reviewer.

Issue #352, opened 2018-03-23
Change flashing to dots


samuelli:
- Remove flashing
 - Favicon is still tied to new, actionable items
 - Add dots for every item that's data changed
 - Dots remain for 10 minutes

![image](https://user-images.githubusercontent.com/787668/37805042-057b5270-2df5-11e8-8ea2-15d3dd4bafe4.png)

samuelli:
A potential change in how this would work. I think this would make it more predictable and more extensible across the board.

The dot marks new activity since you've last read the item. To do this, it'll require us tracking when each user interacted with an item:
 - (to transition from first load/before we track) if there is no last read timestamp - mark last read as now. This means we won't place dots on anything if this is the first time visiting the dashboard or we just added the feature.
 - if latest activity > last read timestamp, show a dot
 - if notification is clicked, sw notifies server that gql object id is read by user x. Server marks gql object keyed off the user id as read.
 - if the item is clicked in the UI, we also fire an event to the server. Result is same as above.
 - locally (on the client), we remove the dot once we fire an event to the server.
 - on a full update - we'll have enough information to sync the dots (should have no change). Locally stored info (can be stored in DOM) about dots is cleared each full update.
 - items always set a latest activity timestamp field - for pull requests this will simply be the latest event timestamp that we are also sending. For issues, this could be the timestamp of the latest comment (TBD).

gauntface:
Testing new activity.

PR #384, opened 2018-04-02
Adds hasNewActivity to API for pull requests

gauntface:
More steps towards: #352

This hooks up the featureEnabled API and adds `hasNewActivity` to the PullRequests API.

Will follow up with the assigned issues next and then add the UI, let me know if you'd rather have these features added immediately.

### DIFF
--- a/src/server/apis/dash-data.ts
+++ b/src/server/apis/dash-data.ts
@@ -5,6 +5,9 @@ import * as api from '../../types/api';
 import {IncomingPullRequestsQuery, mentionedFieldsFragment, prFieldsFragment, PullRequestReviewState, reviewFieldsFragment} from '../../types/gql-types';
 import {github} from '../../utils/github';
 import {userModel} from '../models/userModel';
+import {getPRLastActivityTimestamp} from '../utils/get-pr-last-activity';
+import {issueHasNewActivity} from '../utils/issue-has-new-activity';
+
 import {fetchOutgoingData} from './dash-data/fetch-outgoing-data';
 
 /**
@@ -98,6 +101,8 @@ export async function fetchIncomingData(
       mentioned.set(item.id, mention);
     }
   }
+
+  const lastViewedInfo = await userModel.getAllLastViewedInfo(dashboardLogin);
 
   // Incoming PRs that I've reviewed.
   for (const pr of viewerPrsResult.data.reviewed.nodes || []) {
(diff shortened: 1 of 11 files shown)


src/server/apis/dash-data.ts
```diff
@@ -197,6 +202,15 @@ export async function fetchIncomingData(
       reviewedPr.status = {type: 'ApprovalRequired'};
     }
 
+    const lastActivity = await getPRLastActivityTimestamp(reviewedPr);
+    if (lastActivity) {
+      reviewedPr.hasNewActivity = await issueHasNewActivity(
+          dashboardLogin, lastActivity, lastViewedInfo[pr.id]);
+    } else {
+      reviewedPr.hasNewActivity = false;
```
samuelli: I think its set to false by default below.

src/server/utils/issue-has-new-activity.ts
```diff
@@ -0,0 +1,27 @@
...
+        username, 'featureLastViewed', lastviewedFeature);
+  }
+
+  // We need to account for when the feature was first used and when
+  // the user last interacted with the issue
+  if (!userLastViewed || userLastViewed < lastviewedFeature.enabledAt) {
+    userLastViewed = lastviewedFeature.enabledAt;
+  }
+
+  return (lastActivity > userLastViewed);
```
samuelli: nit: remove parens

src/server/utils/issue-has-new-activity.ts
```diff
@@ -0,0 +1,27 @@
+import {userModel} from '../models/userModel';
+
+export async function issueHasNewActivity(
+    username: string, lastActivity: number, userLastViewed: number|null):
+    Promise<boolean> {
+  const userRecord = await userModel.getUserRecord(username);
```
samuelli: I'm not familiar with firestore and how its caching works. I assume this always does a read.

In that case, I would prefer `userRecord` to be passed in to avoid multiple reads of the same data. The logic to set up the feature can move into `userModel` or another module.

src/server/apis/dash-data.ts
```diff
@@ -197,6 +202,15 @@ export async function fetchIncomingData(
       reviewedPr.status = {type: 'ApprovalRequired'};
     }
 
+    const lastActivity = await getPRLastActivityTimestamp(reviewedPr);
+    if (lastActivity) {
+      reviewedPr.hasNewActivity = await issueHasNewActivity(
+          dashboardLogin, lastActivity, lastViewedInfo[pr.id]);
```
samuelli: This is in the loop, so don't want to be doing more reads than necessary. See comment in `issuesHasNewActivity()`.

gauntface:
after this I would actually like to pull out incoming and make it such that we can share some of the logic for pull requests. This PR contains a couple of pieces of duplicate code and they should be in one place but current structure just forces lots of utils functions to use in both places.

PR #390, opened 2018-04-06
Adding new acitivity indicator

gauntface:
- Adds UI for "has new activity".
- Fixes a bug in user model where some values were needlessly getting reset.

screenshot from 2018-04-06 09-56-03

### DIFF
--- a/gulp-tasks/css.js
+++ b/gulp-tasks/css.js
@@ -18,6 +18,7 @@ function css() {
             customProperties: {
               // Allows both fallback and CSS variables to be used
               preserve: true,
+              warnings: false,
             }
           }
         }),
--- a/src/client/public/scripts/components/dashboard-row.ts
+++ b/src/client/public/scripts/components/dashboard-row.ts
@@ -14,6 +14,7 @@ export interface StatusDisplay {
 }
 
 export interface DashboardRowData {
+  id: string;
   status: StatusDisplay;
   createdAt: number;
   author: string;
@@ -22,6 +23,7 @@ export interface DashboardRowData {
   title: string;
   owner: string;
   repo: string;
+  hasNewActivity: boolean;
   classes?: string[];
 }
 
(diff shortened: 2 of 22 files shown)


gulp-tasks/css.js
```diff
@@ -18,6 +18,7 @@ function css() {
             customProperties: {
               // Allows both fallback and CSS variables to be used
               preserve: true,
+              warnings: false,
```
samuelli: yay

src/client/public/scripts/components/dashboard-row.ts
```diff
@@ -62,20 +64,55 @@ export function genericDashboardRowEventTemplate(data: DashboardRowEventData):
...
+  const hasActivityClasses = ['dashboard-row-status__has-activity'];
+  if (data.hasNewActivity) {
+    hasActivityClasses.push('enabled');
+  }
+
+  async function handleRowClick(event: Event) {
+    let hasActivityElement: HTMLElement|null = null;
+
+    let testNode = event.target as HTMLElement;
+    while (testNode) {
```
samuelli: I would prefer:
```
while (testNode && !testnode.classList.contains('dashboard-row-link') {
  testNode = testNode.parentNode;
}

if (testNode) {
   testNode.querySelector(..);
}
```

This will avoid the break.

src/client/public/scripts/components/dashboard-row.ts
```diff
@@ -62,20 +64,55 @@ export function genericDashboardRowEventTemplate(data: DashboardRowEventData):
...
+      if (testNode && testNode.classList.contains('dashboard-row-link')) {
+        hasActivityElement =
+            testNode.querySelector('.dashboard-row-status__has-activity');
+        break;
+      }
+      testNode = testNode.parentNode as HTMLElement;
+    }
+
+    if (hasActivityElement) {
+      hasActivityElement.classList.remove('enabled');
```
samuelli: It may just be easier to just set an attribute on the parent element such as `has-new-activity`. Then use a CSS attribute selector to control whether or not its visible.

```
.dashboard-row-link[has-new-activity] .dashboard-row-status__has-activity {
  display: block;
}
```

src/server/apis/last-viewed.ts
```diff
@@ -0,0 +1,50 @@
...
+import {userModel} from '../models/userModel';
+
+import {createAPIRoute, ResponseDetails} from './api-route';
+
+async function handleUpdateRequest(request: express.Request):
+    Promise<ResponseDetails<GenericStatusResponse>> {
+  const userRecord = await userModel.getUserRecordFromRequest(request);
+  if (!userRecord) {
+    return {
+      statusCode: 400,
```
samuelli: Use `401 Unauthorized` instead.

gauntface: Technically this should never be reached because we check the user is authorized in the login middleware - but will change regardless.

PR #396, opened 2018-04-06
Moving some API's over to a cleaner standard response

gauntface:
A while ago I had created a sort of pseudo set way of  returning JSON where the response would either be `{error: {code: '', messsage: ''}}` or `{data: *}`. It required a lot of boilerplate and wasn't particularly great to use.

I've spent some time fixing this up so it's a little easier to use:
1. I made a PublicAPIRouter and PrivateAPIRouter class. These essentially wrap the express API so we can force types to be used and customise the callbacks (PrivateAPIRouter provides the UserRecord in the callback).
1. A `response-helper` file provides a few ways to return the right data model - `error('code', 'message')`, `data(....)` or `builder().status(300).cookies({...}).data(...).build()`

If this approach seems ok to you we should be able to remove the public / private routes in the dash-server and be able to drop the login middleware (Since PrivateAPIRouter will check for the userRecord and respond with a 401 response if not provided).


src/server/apis/updates.ts
```diff
@@ -3,36 +3,23 @@ import * as fs from 'fs-extra';
...
-      statusCode: 401,
-      error: {
-        code: 'no-login-details',
-        message: 'No login details were found for this request.',
-      }
-    };
-  }
+import {ResponseDetails} from './abstract-api-router';
+import {PrivateAPIRouter} from './private-api-router';
+import * as reseponseHelper from './response-helper';
```
samuelli: reseponseHelper

gauntface:
@samuelli I took another look at this PR and simplified the typing and removed the builder.

src/server/apis/api-router/response-helper.ts
```diff
@@ -0,0 +1,43 @@
...
+            cookies: undefined,
+          },
+          opts);
+
+  return {
+    statusCode: finalOpts.statusCode,
+    cookies: finalOpts.cookies,
+    data,
+  };
+}
```
samuelli: This feel a lot simpler to me! :)

src/server/apis/updates.ts
```diff
@@ -2,37 +2,22 @@ import * as express from 'express';
...
     const userDetails = await userModel.getUserRecord(request.query.login);
     if (!userDetails) {
-      return {
-        statusCode: 400,
-        error: {
-          code: 'bad-login-request',
-          message: 'User wasn\'t found in user model.',
-        },
-      };
+      return reseponseHelper.error(
```
samuelli: FYI this typo never got fixed.

PR #398, opened 2018-04-09
Convert nav to an element

samuelli:
This adds a really basic base element which renders into light DOM. The nav element which has a single property is converted to use this base element.


src/client/public/scripts/components/base-element.ts
```diff
@@ -0,0 +1,18 @@
...
+export abstract class BaseElement extends HTMLElement {
+  constructor() {
+    super();
+  }
+
+  connectedCallback() {
+    this.renderCallback();
+  }
+
+  renderCallback() {
```
cdata: You may want to consider debouncing invocations of `renderCallback` to avoid multiple re-renders within the same turn.

src/client/public/scripts/components/nav-element.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  set user(user: api.DashboardUser|null) {
+    this._user = user;
+    this.renderCallback();
+  }
+
+  _userTemplate(): TemplateResult {
+    if (!this.user) {
+      return html``;
```
gauntface: Shouldn't we return the project health header?

src/client/public/scripts/components/base-element.ts
samuelli: Yep, what's a good way of doing this? I saw you wrote your own microtask thing it seems?

src/client/public/scripts/dash/dash.ts
```diff
@@ -1,15 +1,18 @@
...
 import {FilterController, FilterId, FilterState} from './filter-controller.js';
 import {genericIssueListTemplate} from './issues.js';
 import {LegendItem, legendTemplate} from './legend.js';
-import {navTemplate} from './nav.js';
 import {notificationCenter} from './notification-center.js';
 import {genericPrListTemplate, outgoingPrListTemplate} from './prs.js';
 import {getLoginParam} from './utils/get-data.js';
 
+// Must export the imported custom elements so they don't get compiled out.
+export {NavElement};
```
gauntface: Could we change this for `import '../components/nav-elements..js'`

PR #410, opened 2018-04-09
Change empty message to an element with attributes

samuelli:
This adds a debouncer to the base element. When observed attributes are changed, `renderCallback()` is called, where `observedAttributes()` is defined by the element implementation.

@cdata PTAL at `base-element` changes.

### DIFF
--- a/src/client/public/scripts/components/base-element.ts
+++ b/src/client/public/scripts/components/base-element.ts
@@ -24,10 +24,10 @@ export abstract class BaseElement extends HTMLElement {
   }
 
   connectedCallback() {
-    this.renderCallback();
+    this.requestRender();
   }
 
-  renderCallback() {
+  requestRender() {
     this._debouncer(() => {
       litRender(this.render(), this);
     });
@@ -37,7 +37,7 @@ export abstract class BaseElement extends HTMLElement {
       _name: string,
       _oldValue: string|null,
       _newValue: string|null) {
-    this.renderCallback();
+    this.requestRender();
   }
 
   abstract render(): TemplateResult;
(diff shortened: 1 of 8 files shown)


src/client/public/scripts/components/base-element.ts
```diff
@@ -1,17 +1,44 @@
...
   }
 
   connectedCallback() {
     this.renderCallback();
   }
 
   renderCallback() {
-    render(this.render(), this);
+    this._debouncer(() => {
+      render(this.render(), this);
```
gauntface: could you rename lit's render to `litRender`.

src/client/public/scripts/components/base-element.ts
```diff
@@ -1,17 +1,44 @@
...
+type Debouncer = (task: Function) => Promise<void>;
+
+function createDebouncer(): Debouncer {
+  let queued = false;
+  return async (task: Function) => {
+    if (queued) {
+      return;
+    }
+    queued = true;
+    await Promise.resolve();
```
gauntface: Can you add some comments on why you are await a promise.resolve - I'm assuming to allow microtasks to finish before rendering.

Is there a reason this is preferable over rAF?

src/client/public/scripts/components/base-element.ts
```diff
@@ -1,17 +1,44 @@
...
+    queued = false;
+    task();
+  };
+}
+
 export abstract class BaseElement extends HTMLElement {
+  private _debouncer: Debouncer;
```
cdata: With TypeScript you could even initialize this inline and avoid writing out the constructor.

src/client/public/scripts/components/base-element.ts
cdata: In the microtask case you are avoiding rerender after every render call in a single turn. The raf case also works as long as you are okay with waiting the extra time to render. Microtask is just the minimum viable timing to use for this model IMO.

src/client/public/scripts/components/base-element.ts
samuelli: Yes!

src/client/public/scripts/components/base-element.ts
samuelli: What he said.

src/client/public/scripts/components/base-element.ts
samuelli: ++, done.

PR #416, opened 2018-04-10
Visually center the main content

### LINKED ISSUE #415: Center the main view
Main view is not visually centered. The main view should always be centered regardless of the nav menu.

samuelli:
Fixes #415.

### DIFF
--- a/src/client/public/styles/base.css
+++ b/src/client/public/styles/base.css
@@ -22,6 +22,13 @@ body {
   margin: 0 auto;
 }
 
+@media (max-width: 1550px) {
+  .main-content {
+    margin-left: calc(var(--nav-size) + var(--grid-size));
+    width: auto;
+  }
+}
+
 h1, h2 {
   font-family: var(--heading-font);
 }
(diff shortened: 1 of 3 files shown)


src/client/public/styles/nav.css
```diff
@@ -1,11 +1,16 @@
...
   flex-direction: column;
   margin-right: var(--grid-size);
   font-size: 14px;
   line-height: 20px;
   flex-shrink: 0;
+  position: absolute;
+
+  --grid-size: 16px;
+  --max-desktop-width: 1024px;
+  --nav-size: 250px;
```
gauntface: Stupid question - will these variables be picked up even though they are defined *after* usage?

src/client/public/styles/nav.css
```diff
@@ -1,11 +1,16 @@
 nav-element {
-  width: 250px;
+  width: var(--nav-size);
```
gauntface: --nav-size is defined twice, once in the root and once in the element, would it make sense to use params with a hardcoded fallback? `width: var(--nav-size, 250px);`

src/client/public/styles/nav.css
samuelli: Removed, shouldn't be there.

src/client/public/styles/nav.css
samuelli: Fixed.

PR #417, opened 2018-04-10
PR activity now accounts for user actions

### LINKED ISSUE #412: PR new activity dot can be OTT
If the user is the cause of the latest activity on a Pull Request, the dot should not be shown.

For example - The user approved the changes and therefore shouldn't have the dot.

screenshot from 2018-04-09 15-59-23

gauntface:
Fixes #412


src/test/server/apis/outgoing-prs-test.ts
```diff
@@ -309,14 +316,15 @@ test('[outgoing-prs-test]: with success status', (t) => {
       allow_squash_merge: true,
       allow_merge_commit: true,
     },
-    mergeable: MergeableState.UNKNOWN,
+    mergeable: MergeableState.MERGEABLE,
```
samuelli: Why did this change?

src/types/api.ts
```diff
@@ -35,6 +40,7 @@ export interface PullRequest {
   avatarUrl: string;
   status: PullRequestStatus;
   events: PullRequestEvent[];
+  lastComment: LastComment|null;
```
samuelli: Why has this been added to the API?

samuelli: This should be entirely on the server side, and rolled into the `hasNewActivity` boolean below.

src/test/server/apis/outgoing-prs-test.ts
gauntface: I *believe* it's because we are using an actual github token for project-health1 where as before it was just public information.

gauntface:
Removed from the API.

PR #420, opened 2018-04-11
Prevent FOUC and minor style tweaks

### LINKED ISSUE #383: Popular flames wrap on small screen.
Would be great if they had min width.

screen shot 2018-04-02 at 11 54 01 am

samuelli:
Fixes #383

### DIFF
--- a/src/client/public/styles/components/toggle-element.css
+++ b/src/client/public/styles/components/toggle-element.css
@@ -23,7 +23,3 @@ toggle-element[disabled] {
   opacity: 0.4;
   cursor: not-allowed;
 }
-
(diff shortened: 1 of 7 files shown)


src/client/public/styles/components/toggle-element.css
```diff
@@ -21,3 +23,7 @@ toggle-element[disabled] {
   opacity: 0.4;
   cursor: not-allowed;
 }
+
+toggle-element[selected] .toggle-element__icon {
```
gauntface: Is this needed?

samuelli: No.

PR #421, opened 2018-04-11
Use icon font instead of SVG icons

samuelli:


samuelli:
> This will show the page :(

Probably, I'm not sure I can make it async cause that's what I'd like. Ergonomically, I much prefer this. Maybe we can defer loading of the font somehow.

gauntface:
I've generally always inlined SVG's in a page service side.

This provides `fill` support and is instant to load.

PR #422, opened 2018-04-12
Issue activity

gauntface:
This adds issue activity and improves test coverage for the issue API's.

screenshot from 2018-04-11 17-23-35


src/server/apis/issues.ts
```diff
@@ -1,99 +1,138 @@
...
+        popularity: fetchPopularity(node),
+        hasNewActivity,
+        status: statusCb(node, assigneeLogin),
+      });
     }
+  }
+
+  const issuesResponse: IssuesResponse = {
+    issues: issues.sort((a: Issue, b: Issue) => {
+      if (a.hasNewActivity === b.hasNewActivity) {
```
samuelli: This sorts all the assigned issues too right? Can we instead change this to be sorted by `lastActivity`? This would mean ordering is not based on the dots, but it makes it very likely that dots appear at the top. This will result in the UI not jumping around if you read an item, then refresh etc.

PR #423, opened 2018-04-12
Allow multiple orgs to be specified in metric calculation

samuelli:


src/test/cli/metrics/review-latency-test.ts
```diff
@@ -20,15 +20,15 @@ test.afterEach.always(async (t) => {
...
+  const result = await getReviewLatency({org: ['WebComponents'], since});
   t.is(result.totalLatency, 262575635000);
   t.is(result.reviews.length, 684);
 });
 
 test.serial('WebComponents/webcomponents.org review latency', async (t) => {
   const since = new Date(0);
   const result = await getReviewLatency(
-      {org: 'WebComponents', repo: 'webcomponents.org', since});
+      {org: ['WebComponents'], repo: 'webcomponents.org', since});
```
gauntface: I don't understand why this is an array. The repo (in most cases) will long apply to one org

samuelli: I think ideally the pulling of repos is not part of these metric classes and then the metric is calculated separately.

Here, review latency expects an array for the org argument. This is sufficient for what we're using the cli for right now.

PR #424, opened 2018-04-12
Fixing FF support for CustomElements

gauntface:
Firefox 60 supports modules but doesn't seem to work with custom elements.

Removing the nomodule option gets everything back up and running in FF.


gauntface:
This works in Chrome and we block on a font for icons, so I really don't see this as an issue to add browser support.

samuelli:
> This works in Chrome and we block on a font for icons, so I really don't see this as an issue to add browser support.

I'm unclear on what the script does on Chrome, but we can fix it later.

PR #427, opened 2018-04-12
Factor out a row-header element

samuelli:


src/client/public/scripts/components/row-header.ts
```diff
@@ -0,0 +1,102 @@
...
+    return this._extraHeaderData;
+  }
+
+  set extraHeaderData(data: TemplateResult[]|undefined) {
+    this._extraHeaderData = data;
+    this.requestRender();
+  }
+
+  render() {
+    async function handleRowClick(event: Event) {
```
gauntface: This is going to force a mixture of logic with view, is that normal in WC?

samuelli: Yes, but also this part is identical to before.

PR #432, opened 2018-04-13
Adding tests and fixes for API route

gauntface:


src/test/server/apis/api-router/private-router-test.ts
```diff
@@ -0,0 +1,126 @@
...
+  t.context = {
+    sandbox: sinon.sandbox.create(),
+  };
+});
+
+test.afterEach.always(async (t) => {
+  t.context.sandbox.restore();
+});
+
+test.serial(
```
samuelli: AFAIK, this doesn't need to run in serial. Don't see any singletons that you're potentially using.

src/test/server/apis/api-router/private-router-test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      const callbackSpy = t.context.sandbox.spy(async () => {
+        return responseHelper.data({
+          hello: 'world',
+        });
+      });
+      const privateRouter = new PrivateAPIRouter();
+      privateRouter.post('/', callbackSpy);
+
+      t.deepEqual(postSpy.callCount, 1, 'express.post() call count');
```
samuelli: Maybe use `equal` for non objects?

src/test/server/apis/api-router/private-router-test.ts
gauntface: It's the stubbing that makes this need to be serial

PR #434, opened 2018-04-13
Convert rows to elements

samuelli:


src/client/public/scripts/components/row-element.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export class RowElement extends BaseElement {
+  @property() data: RowData|null = null;
+  @property() extraHeaderData: TemplateResult[] = [];
+  @property() events: RowEvent[] = [];
+
+  _renderEvent(event: RowEvent) {
+    const timeTemplate = (time: number) =>
+        html`<time class="dashboard-row-event__time" datetime="${
+            new Date(time).toISOString()}">${timeToString(time)}</time>`;
```
gauntface: Is there a need for lit-html here vs just passing the template literal?

src/client/public/scripts/components/row-element.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    const linkTemplate = (url: string, text: string|TemplateResult) =>
+        html`<a class="dashboard-row-event__url" href="${
+            url}" target="_blank">${text}</a>`;
+
+    return html`
+<div class$="dashboard-row-event ${
+        event.classes ? event.classes.join(' ') : ''}">
+  ${
+        event.time ? timeTemplate(event.time) :
```
gauntface: Could you move this if statement into the timeTemplate function?

src/client/public/scripts/components/row-header.ts
```diff
@@ -3,12 +3,11 @@ import {TemplateResult} from '../../../../../node_modules/lit-html/lit-html.js';
...
 
 import {BaseElement, property} from './base-element.js';
-import {DashboardRowData} from './dashboard-row.js';
+import {RowData} from './row-element.js';
 
 export class RowHeader extends BaseElement {
-  @property({attribute: true}) rowData: DashboardRowData|undefined = undefined;
-  @property({attribute: true})
-  extraHeaderData: TemplateResult[]|undefined = undefined;
+  @property() rowData: RowData|undefined = undefined;
```
gauntface: Aren't these being used as attributes in the row-element?

samuelli: Nope & shouldn't be.

src/client/public/scripts/components/row-element.ts
samuelli: Yes, it must be a lit tagged template literal, otherwise we will end up with it as a string in the DOM instead of these being converted to elements.

src/client/public/scripts/components/row-element.ts
samuelli: Done.

PR #436, opened 2018-04-13
Implement basic router and app page switcher

samuelli:
This is currently unused since none of the pages are elements. Since `dash.ts` is huge, I'm pulling conversion of that into a separate PR.

### DIFF
--- a/src/client/public/scripts/app/app.ts
+++ b/src/client/public/scripts/app/app.ts
@@ -1,0 +1,60 @@
+import './router.js';
+
+/**
+ * Manages pages within the app.
+ */
+class AppElement extends HTMLElement {
+  private modules: string[] = [];
+  private loadedModules: string[] = [];
+
+  connectedCallback() {
+    document.body.addEventListener('url-changed', this.updatePage.bind(this));
+    for (const element of Array.from(this.children)) {
+      this.modules.push(element.id);
+    }
+
+    // Initial load.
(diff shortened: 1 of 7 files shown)


src/client/public/scripts/app/router.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Fire an event if there are same origin changes.
+ */
+function globalClickHandler(event: MouseEvent) {
+  if (event.defaultPrevented) {
+    return;
+  }
+
+  const path = getSameOriginLink(event);
```
gauntface: Maybe rename this to isSameOriginLink or filterSameOriginLink.

Took me a while to figure out what this was for.

src/client/public/scripts/app/app.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+   * Load the correct page based on the path.
+   */
+  updatePage() {
+    const path = window.location.pathname;
+    let module = 'dash';
+    const normalized = path.substring(1);
+    if (this.modules.includes(normalized)) {
```
gauntface: This normalized piece is a little confusing to following, could you add some commebts

src/client/public/scripts/app/router.ts
samuelli: Done

PR #437, opened 2018-04-14
Convert dashboard to dash-page element

samuelli:


samuelli:
Apologies for the large PR @gauntface. I think there's still a lot to clean up to get to a really good spot. I do think some Redux work may help us separate out things like polling from the UI and we can have saner properties which we bind rendering to instead. All that pending on some Redux investigation.

gauntface:
Maybe it's worth taking a step back and planning out our approach. At the moment we seem to be adding more and more layers of indirection

PR #438, opened 2018-04-16
Adding github app install event

gauntface:


src/server/controllers/webhook-events/github-app-install.ts
```diff
@@ -0,0 +1,72 @@
...
+import {githubAppModel} from '../../models/githubAppModel';
+
+export interface InstallHook {
+  action: 'created';
+  installation: {
+    id: number; repository_selection: 'all' | 'selected';
+    permissions: {[name: string]: string;};
+    events: string[];
+    account: {login: string; avatar_url: string; type: 'User' | 'Organization';}
+    repositories: Array<{id: number; name: string; full_name: string}>
```
samuelli: Should be `,` between all these lines, and in all these objects. Only the top level `interface` uses `;`.

src/server/controllers/webhook-events/github-app-install.ts
```diff
@@ -0,0 +1,72 @@
...
+  for (const repo of hookBody.installation.repositories) {
+    const repoId = await github().query<RepoIdQuery>({
+      query: repoIdQuery,
+      variables: {
+        owner: hookBody.installation.account.login,
+        name: repo.name,
+      },
+      fetchPolicy: 'network-only',
+      context: {
+        token: tokens['github-app-to-graphql'],
```
samuelli: What is this?

src/server/controllers/webhook-events/github-app-install.ts
```diff
@@ -0,0 +1,72 @@
...
+    repository_selection: hookBody.installation.repository_selection,
+    type: hookBody.installation.account.type,
+    login: hookBody.installation.account.login,
+    avatar_url: hookBody.installation.account.avatar_url,
+  });
+
+  const tokens = await fse.readJson(
+      path.join(__dirname, '..', '..', '..', '..', 'tokens.json'));
+
+  for (const repo of hookBody.installation.repositories) {
```
samuelli: Does this return the full list if `all` is selected?

gauntface: yes it does

PR #450, opened 2018-04-18
Dash data private api

gauntface:
Whatever happened last night, there were a few things that could have been better:

1. The errors shouldn't have been uncaught. This PR moved the dash data to the private api router which will catch errors.
1. I pulled out the incoming PR into a seperate file.
1. I've tried to pull out GitHub's error instead of letting Apollo throw a network error.

### DIFF
--- a/src/server/apis/dash-data/handle-incoming-pr-request.ts
+++ b/src/server/apis/dash-data/handle-incoming-pr-request.ts
@@ -1,0 +1,371 @@
+import * as express from 'express';
+
+import * as api from '../../../types/api';
+import {IncomingPullRequestsQuery, mentionedFieldsFragment, PullRequestReviewState, reviewFieldsFragment} from '../../../types/gql-types';
+import {github} from '../../../utils/github';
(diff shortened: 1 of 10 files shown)


src/server/apis/dash-data/handle-incoming-pr-request.ts
```diff
@@ -0,0 +1,371 @@
...
+      continue;
+    }
+
+    // Find relevant review.
+    let relevantReview: MyReviewFields|null = null;
+    if (pr.reviews && pr.reviews.nodes) {
+      // Generated gql-types is missing the proper __type field so a cast is
+      // needed.
+      relevantReview =
+          findMyRelevantReview(pr.reviews.nodes as Array<MyReviewFields|null>);
```
samuelli: This cast may no longer be necessary since I saw that gql-types has been changing.

gauntface: It complains if I remove it so will leave it in for now.

PR #451, opened 2018-04-18
Generate and store top ranked repos by user

samuelli:
This change adds an `/api/user` endpoint which returns all of the user's information required to populate the nav. On the first hit to this endpoint, it will populate repos from the `my-repos` script which pulls metrics and uses a heuristic to determine top ranked repos.

This change also removes `clang-format` from our dependency list, forcing local versions of `clang-format` for more consistency. It also renames `tsconfig-node.json` to `tsconfig.json` in the top level, which allows editors to properly recognise the TS configuration.


samuelli:
@aomarks: PTAL at cursored query changes.

gauntface:
@samuelli if you use the local `clang-format, won't that cause more inconsistency as we'll both have different versions.

src/server/apis/user.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Fetches the user profile of the authenticated user. Emulation of another user
+ * is forbidden here.
+ */
+async function handlerUser(
+    _request: express.Request, userRecord: UserRecord): Promise<APIResponse> {
+  let repos = userRecord.repos;
+  if (!repos) {
+    repos = await getMyRepos(userRecord.username, userRecord.githubToken);
+    userModel.updateRepos(userRecord.username, repos);
```
gauntface: should you await this?

src/server/apis/user.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import {APIResponse} from './api-router/abstract-api-router';
+import {PrivateAPIRouter} from './api-router/private-api-router';
+import * as responseHelper from './api-router/response-helper';
+
+/**
+ * Fetches the user profile of the authenticated user. Emulation of another user
+ * is forbidden here.
+ */
+async function handlerUser(
```
gauntface: Maybe rename to `handleUserRequest`

src/server/utils/my-repos.ts
```diff
@@ -25,66 +26,89 @@ const MAX_STATS_RETRIES = 5;
...
 // scale of the resulting scores.
 const CONTRIBUTION_SIZE = 10;
-const CONTRIBUTION_WINDOW = 1000 * 60 * 60 * 24 * 90;  // 90 days.
+const CONTRIBUTION_WINDOW = 1000 * 60 * 60 * 24 * 7;  // 7 days.
 
 /**
  * Returns a list of repos that the given user contributes to. It is ordered
  * based on contribution recency and consistency.
  */
 export async function getMyRepos(
```
gauntface: I'd rename this to `getAutomatedRepoList`, just to differentiate between the stored repo's vs the first load of repos.

aomarks:
> @aomarks: PTAL at cursored query changes.

+1 looks nicer using a WatchQueryOptions object instead of positional args

samuelli:
> @samuelli if you use the local `clang-format, won't that cause more inconsistency as we'll both have different versions.

Actually maybe it's just out of date. Updated to use NPM one, seems to not be causing the same issue as before.

src/server/apis/user.ts
samuelli: Done.

src/server/apis/user.ts
samuelli: Done.

src/server/utils/my-repos.ts
samuelli: Done.

PR #458, opened 2018-04-19
Add repos to the nav

samuelli:


src/client/public/scripts/components/nav-element.ts
```diff
@@ -5,38 +5,62 @@ import * as api from '../../../../types/api.js';
...
+    return html`
+      <a class="nav-item selected" href="/">
+        <img class="nav-item__avatar" src="${this.data.avatarUrl}"
+            alt="Avatar of ${this.data.login}" />
+        <div class="nav-item__name">${this.data.login}</div>
+      </a>
+    `;
+  }
 
+  _header(): TemplateResult {
```
gauntface: Shouldn't these just be marked as private?

PR #460, opened 2018-04-19
Adding editor for org settings

gauntface:


src/client/private-pages/index.html
```diff
@@ -19,6 +20,7 @@
   <app-element>
     <dash-page id="dash"></dash-page>
     <dash-pagination id="all"></dash-pagination>
+    <github-app-config id="github-app-config"></github-app-config>
```
samuelli: `/org/config/:orgname` would be my preference. Redirect on GitHub install from installation id => org config URL.

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
...
+import {TemplateResult} from '../../../../../node_modules/lit-html/lit-html.js';
+import * as api from '../../../../types/api.js';
+import {BaseElement, property} from '../components/base-element.js';
+
+const DEFAULT_CONFIG = `{
+  // Add custom configuration here
+}`;
+
+class GithubAppConfig extends BaseElement {
+  private orgName: string|null;
```
samuelli: This would probably change to a property as well once the URL changes are made.

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
...
+    this.initialised = true;
+  }
+
+  async getConfig(): Promise<string|null> {
+    const response = await fetch('/api/github-app/get-config', {
+      method: 'POST',
+      headers: {'Content-Type': 'application/json'},
+      body: JSON.stringify({
+        orgName: this.orgName,
+      }),
```
samuelli: It seems like this should be GET.

`GET /api/org/config/:orgname`.

Then to modify it:

`POST /api/org/config/:orgname`.

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
...
+      credentials: 'include',
+    });
+
+    // tslint:disable-next-line:no-any
```
samuelli: You're not using any - can remove this.

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
...
+    return responseBody.data.fileContents;
+  }
+
+  async getOrgNameFromInstallId(installId: string): Promise<string|null> {
+    const response = await fetch('/api/github-app/details', {
+      method: 'POST',
+      headers: {'Content-Type': 'application/json'},
+      body: JSON.stringify({
+        installId: Number(installId),
+      }),
```
samuelli: Again, I think I would rather this be a GET since its not modifying anything.

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
...
+      return null;
+    }
+
+    const installData = responseBody.data;
+    return installData.login;
+  }
+
+  render(): TemplateResult {
+    if (this.errorMsg) {
+      return html`<h2>GitHub App Configuration</h2>
```
samuelli: Maybe `:org settings`

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    let configDisplay = DEFAULT_CONFIG;
+    if (this.currentConfig) {
+      configDisplay = this.currentConfig;
+    }
+
+    return html`
+<h2>GitHub App Configuration</h2>
+<div class="editor-container">
+<textarea class="github-app-defaults text-editor">
```
samuelli: `<div>` should be indented.

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
...
+  "issues.priorityIssues": false,
+}
+</textarea>
+<textarea id="settings-text-area" class="text-editor">
+${configDisplay}
+</textarea>
+</div>
+<div class="button-container">
+  <button id="reset-btn" on-click="${() => this.resetConfig()}">Reset</button>
+  <button id="save-btn" on-click="${() => this.saveConfig()}">Save</button>
```
samuelli: I believe this can be `${this.saveConfig}`. You may have to add `.bind(this)`.

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
...
+</div>
+`;
+  }
+
+  private async saveConfig() {
+    const settingsTextArea = document.querySelector('#settings-text-area');
+    const resetButton = document.querySelector('#reset-btn');
```
samuelli: For simplification, can we remove the reset functionality?

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
...
+</div>
+<div class="button-container">
+  <button id="reset-btn" on-click="${() => this.resetConfig()}">Reset</button>
+  <button id="save-btn" on-click="${() => this.saveConfig()}">Save</button>
+</div>
+`;
+  }
+
+  private async saveConfig() {
+    const settingsTextArea = document.querySelector('#settings-text-area');
```
samuelli: `this.querySelector()`.

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
...
+    if (!saveButton) {
+      console.warn('Unable to find the \'#save-btn\' element.');
+      return;
+    }
+
+    settingsTextArea.setAttribute('disabled', 'true');
+    resetButton.setAttribute('disabled', 'true');
+    saveButton.setAttribute('disabled', 'true');
+
+    const response = await fetch('/api/github-app/save-config', {
```
samuelli: Yeah, I would probably decouple this from `github-app`. As above, I would prefer this to be `POST /api/org/:org`. Orgs have configs which we'll always keep around. The GitHub app will read this config, but they could conceivably be separate. eg. Maybe I re-install the app - it should keep the config though.

src/client/public/scripts/pages/github-app-config.ts
```diff
@@ -0,0 +1,202 @@
+import {html} from '../../../../../node_modules/lit-html/lib/lit-extended.js';
+import {TemplateResult} from '../../../../../node_modules/lit-html/lit-html.js';
+import * as api from '../../../../types/api.js';
+import {BaseElement, property} from '../components/base-element.js';
+
+const DEFAULT_CONFIG = `{
+  // Add custom configuration here
+}`;
+
+class GithubAppConfig extends BaseElement {
```
samuelli: I would probably rename this. See below comments on separating the app a little from org config.

src/types/api.ts
```diff
@@ -97,6 +101,35 @@ export interface UserResponse {
...
+  };
+}
+
+export interface JSONAPIDataResponse<T> {
+  data: T;
+}
+
+export type JSONAPIResponse<T> = JSONAPIErrorResponse|JSONAPIDataResponse<T>;
+
+export interface GithubAppInstall {
```
samuelli: This doesn't seem like it belongs in the API. From what I can see, this is used in the API to convert from install id to org name. This would then just need to be a single string `org: string`.

src/types/api.ts
```diff
@@ -97,6 +101,35 @@ export interface UserResponse {
   repos: Repository[];
 }
 
+export interface OrgSettings {
+  fileContents: string;
+  lastUpdated: number;
+  editors: string[];
```
samuelli: What's this?

src/server/apis/github-app.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  return responseHelper.data<GenericStatusResponse>({
+    status: 'ok',
+  });
+}
+
+export function getRouter(): express.Router {
+  const appRouter = new PrivateAPIRouter();
+  appRouter.post('/details', handleDetailsRequest, {
```
samuelli: As above, some of these move to GET and use [route parameters](https://expressjs.com/en/api.html#req.params) instead. `:org`.

src/server/models/settingsModel.ts
```diff
@@ -0,0 +1,52 @@
...
+      fileContents: newSettings,
+      lastUpdated: Date.now(),
+      editors: editorsList,
+    };
+
+    await orgDocRef.set(newConfig);
+  }
+}
+
+export const settingsModel = new SettingsModel();
```
samuelli: This needs some basic tests.

src/types/api.ts
gauntface: the users who edit the settings.

gauntface:
@samuelli I've added tests, added an endpoint to redirect post-install > org config page and updated for latest changes.

src/server/apis/org-config.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  if (typeof newSettings !== 'string') {
+    return responseHelper.error(
+        'settings-not-a-string',
+        'This API expects the \'settings\' api to be a plain text string.');
+  }
+
+  try {
+    await settingsModel.setOrgSettings(
```
samuelli: We need to do authorization checks. Currently this API would let anyone with a valid login to modify any org's settings. Instead, we'll need to do a GitHub API authorization check before each read/write.

gauntface:
Should be good for another round of reviews.

src/client/public/scripts/pages/org/config.ts
```diff
@@ -0,0 +1,170 @@
...
+    this.initialised = true;
+  }
+
+  async getConfig(): Promise<string|null> {
+    const response = await fetch(`/api/org/config/${this.orgName}`, {
+      method: 'GET',
+      credentials: 'include',
+    });
+
+    // tslint:disable-next-line:no-any
```
samuelli: You're not using `any` here.

src/server/models/settingsModel.ts
```diff
@@ -0,0 +1,102 @@
...
+                          .collection(ORG_SETTINGS_COLLECTION_NAME)
+                          .doc(orgOrUser);
+
+    const snapshot = await orgDocRef.get();
+    if (!snapshot.exists) {
+      return null;
+    }
+
+    return snapshot.data() as OrgSettings;
+  }
```
samuelli: nit: new line

src/server/utils/generate-github-app-token.ts
```diff
@@ -0,0 +1,43 @@
...
+  );
+
+  const response = await fetch(
+      `https://api.github.com/installations/${installId}/access_tokens`, {
+        method: 'POST',
+        headers: {
+          'Accept': 'application/vnd.github.machine-man-preview+json',
+          'Authorization': `Bearer ${jwt}`,
+        },
+      });
```
samuelli: Why don't we use the `github()` module here?

PR #462, opened 2018-04-20
Add repo page with untriaged issues

samuelli:


src/client/public/scripts/app/app.ts
```diff
@@ -24,24 +27,30 @@ class AppElement extends HTMLElement {
...
   /**
    * Switches the visible page.
    */
-  setVisiblePage(module: string) {
+  setVisiblePage(selectedModule: string, path: string) {
     for (const element of Array.from(this.children)) {
-      if (element.id === module) {
+      if (element.getAttribute('prefix') === selectedModule) {
         element.setAttribute('selected', '');
+        element.setAttribute('path', path);
```
gauntface: I'm kind of getting lost with prefix and path. I understand path may have additional details beyond the prefix, but why are you setting it as an attribute on the element?

src/client/public/scripts/components/nav-element.ts
```diff
@@ -6,20 +6,29 @@ import {BaseElement, property} from './base-element.js';
 
 export class NavElement extends BaseElement {
   @property() data: api.UserResponse|null = null;
+  @property() documentUrl = document.location.pathname;
```
gauntface: Add a comment here explaining this is just to take advantage of property binding.

src/client/public/scripts/components/nav-element.ts
```diff
@@ -6,20 +6,29 @@ import {BaseElement, property} from './base-element.js';
...
 
   async connectedCallback() {
     const response = await fetch('/api/user', {credentials: 'include'});
     // TODO: change this to use the JSON API response
     this.data = (await response.json()).data as api.UserResponse;
+
+    document.body.addEventListener('url-changed', this.urlChanged.bind(this));
```
gauntface: I feel like you could just inline this method as a fat arrow function.

src/client/public/scripts/components/nav-element.ts
```diff
@@ -6,20 +6,29 @@ import {BaseElement, property} from './base-element.js';
...
+  private urlChanged() {
+    this.documentUrl = document.location.pathname;
   }
 
   private userTemplate(): TemplateResult {
     if (!this.data) {
       return html``;
     }
 
+    const selected = this.documentUrl === '/' ? 'selected' : '';
```
gauntface: The visible page is set based on a "prefix" value. These selected items are set based on URL.

Could we instead have a single prefix value that the app-element uses to show a visible page and then give the value to the nav-element as a property / attribute. From this, can we render the elements and then query for an element with attribute `document.querySelector('*[prefix=${selectedPrefix}]')`. Then just set selected on that element.

src/client/public/scripts/pages/repo.ts
```diff
@@ -0,0 +1,86 @@
...
+import {BaseElement, property} from '../components/base-element.js';
+import {FilterLegendEvent, FilterLegendItem} from '../components/filter-legend.js';
+import {filterController, FilterId} from '../dash/filter-controller.js';
+import {genericIssueListTemplate} from '../dash/issues.js';
+
+class RepoPage extends BaseElement {
+  @property({attribute: true}) path = '';
+  @property() untriaged: api.Issue[] = [];
+  private cachedId = '';
+  private filters: {[key: string]: [FilterLegendItem]} = {
```
gauntface: Does this need a filter given it's a single element?

src/client/public/scripts/app/app.ts
```diff
@@ -11,7 +11,10 @@ class AppElement extends HTMLElement {
   connectedCallback() {
     document.body.addEventListener('url-changed', this.updatePage.bind(this));
     for (const element of Array.from(this.children)) {
-      this.modules.push(element.id);
+      const prefix = element.getAttribute('prefix');
+      if (prefix) {
+        this.modules.push(prefix);
+      }
```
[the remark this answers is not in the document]
samuelli: Done

src/client/public/scripts/app/app.ts
samuelli: Ack, changed to `urlParams`.

src/client/public/scripts/components/nav-element.ts
samuelli: Done.

src/client/public/scripts/components/nav-element.ts
samuelli: Done.

src/client/public/scripts/components/nav-element.ts
samuelli: I think we synced offline on this.

src/client/public/scripts/pages/repo.ts
samuelli: Gonna leave this in. I think I might factor out a generic concept of a filtered list.

PR #465, opened 2018-04-23
Add remove repo from menu functionality

samuelli:
When the selected menu item is hovered on, buttons will appear. The remove button will permanently remove the repo from your menu.

### DIFF
--- a/src/server/apis/user.ts
+++ b/src/server/apis/user.ts
@@ -33,29 +33,33 @@ export async function handleRemoveRepo(
   const owner = request.body.owner;
   const name = request.body.name;
 
-  if (!owner || !name) {
-    return responseHelper.error(
-        'invalid-request', 'Missing owner/name from request');
+  if (!owner) {
+    return responseHelper.error('invalid-request', 'Missing owner in request');
   }
 
-  const oldRepos = userRecord.repos;
-  const newRepos = [];
-  if (!oldRepos) {
+  if (!name) {
+    return responseHelper.error('invalid-request', 'Missing name in request');
+  }
+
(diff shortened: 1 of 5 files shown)


src/server/apis/user.ts
```diff
@@ -28,8 +28,40 @@ async function handlerUserRequest(
   });
 }
 
+export async function handleRemoveRepo(
+    request: express.Request, userRecord: UserRecord): Promise<APIResponse> {
+  const owner = request.body.owner;
+  const name = request.body.name;
+
+  if (!owner || !name) {
```
gauntface: I tend to avoid these kinds of checks, while it's easier to write, the natural first question when something goes wrong is - am I missing an owner or am I missing a name.

src/server/apis/user.ts
```diff
@@ -28,8 +28,40 @@ async function handlerUserRequest(
...
+
+  const oldRepos = userRecord.repos;
+  const newRepos = [];
+  if (!oldRepos) {
+    return responseHelper.error('no-repos', 'No repos found on user');
+  }
+
+  for (const repo of oldRepos) {
+    if (repo.owner !== owner || repo.name !== name) {
+      newRepos.push(repo);
```
gauntface: Could you change this to splice out the one repo from the array and break when it's found?

src/server/apis/user.ts
samuelli: Fixed

src/server/apis/user.ts
samuelli: Done.

PR #480, opened 2018-04-25
Adding a settings models

gauntface:
Idea is - when a settings file is changed (at the moment the org config file is changed / updated) calling `installationSettingsChange()` will step through an orgs repo's and setup any settings accordingly

This works by running `initSettings()` is needed which will import all files under `server/controllers/settings/` which will add themselves as settings, passing in a key and an instance of a `Setting` class.

After this, `installationSettingsChange()` will look at the users config and call `run()` on any setting it is aware of.

At the moment I'm using `any` for the actual setting as a user can define anything they want, but perhaps we should be using what we expect in the Setting panel?

For example, this will currently save:

```
{
  "issues.priorityIssues": {
    "value": true
  }
}
```

But the expected value is:

```
{
  "issues.priorityIssues": true
}
```

I'm not sure what is best practice for this / how we should support it.

### DIFF
--- a/src/server/controllers/github-app-settings.ts
+++ b/src/server/controllers/github-app-settings.ts
@@ -1,74 +1,72 @@
 import * as fse from 'fs-extra';
-import {validate} from 'jsonschema';
+import {validate as validateAgainstSchema} from 'jsonschema';
 import * as path from 'path';
 
 import {githubAppModel, GithubRepo} from '../models/githubAppModel';
 
-// TODO: How should this be configured
+// TODO: This might be better of as conditional types
 type AvailableIssueTypes = boolean;
 
-interface SettingsModuleConfig {
+interface AppPluginConfig {
   default: AvailableIssueTypes;
   description: string;
   type: string;
 }
 
-export interface SettingsModule {
-  config(): {[key: string]: SettingsModuleConfig};
+export interface AppPlugin {
+  config(): {[key: string]: AppPluginConfig};
 
-  // tslint:disable-next-line:no-any
-  run: (settingConfig: any, repos: GithubRepo[]) => Promise<void>;
+  settingsChanged<T>(settingConfig: T, repos: GithubRepo[]): Promise<void>;
 }
 
 class Settings {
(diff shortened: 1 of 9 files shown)


src/server/controllers/settings/priority-issues.ts
```diff
@@ -0,0 +1,11 @@
...
+import {addSetting, Setting} from '../github-app-settings';
+
+class PriorityIssueSetting implements Setting {
+  // tslint:disable-next-line:no-any
+  async run(setting: any, repos: GithubRepo[]) {
+    console.log('Priority Issue.: ', setting, 'on: ', repos);
+  }
+}
+
+addSetting('issues.priorityIssues', new PriorityIssueSetting());
```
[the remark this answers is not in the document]
gauntface: If the developer adds two keys, should the settings module be called once or twice?

I.e. for:

```
{
  "example/1": {
    default: true,
    description: "This does blah",
    type: "boolean",
  },
  "example/2": {
    default: true,
    description: "This does blah",
    type: "boolean",
  }
}
```

src/server/controllers/github-app-settings.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.settingsModules = [];
+    this.moduleKeys = new Set();
+
+    this.loadAppPlugins();
+  }
+
+  loadAppPlugins() {
+    const settingsDir = path.join(__dirname, 'settings');
+
+    // Autoloads priority issues
```
samuelli: Remove this comment.

src/server/controllers/settings/priority-issues.ts
```diff
@@ -0,0 +1,27 @@
+import {GithubRepo} from '../../models/githubAppModel';
+import {AppPlugin, settings} from '../github-app-settings';
+
+// TODO: Conditional types might allow us to define this from config()
+export interface PluginSetting {
+  'issues.priorityIssues': boolean;
+}
+
+class PriorityIssueSetting implements AppPlugin {
```
samuelli: Rename this class

PR #489, opened 2018-04-27
Don't allow double registration of custom elements

### LINKED ISSUE #484: Firefox custom elements are broken
We should use `customElements.get()` before registering elements to avoid multiple defines.

samuelli:
Fixes #484


abdonrd:
@samuelli @gauntface Why can it be the case that you register the same element several times?

PR #498, opened 2018-04-30
Add production app

samuelli:


src/utils/secrets.ts
```diff
@@ -21,5 +24,14 @@ export function initSecrets(secrets: ApplicationSecrets) {
   if (secretsSingleton) {
     throw new Error('Secrets already initialised.');
   }
+  if (!secrets.APP && process.env.GAE_APPLICATION === 'github-health') {
```
gauntface: Would be nice if production flag was actually something like 'process.env.NODE_ENV === 'production' | 'staging''

src/server/controllers/webhook-events/github-app-install.ts
```diff
@@ -73,7 +73,7 @@ async function handleNewAppInstall(hookBody: InstallHook) {
       // itself. At the moment this is using a persona access token
       // to get around restriction of GitHub app not having access
       // to GQL.
-      token: secrets().GITHUB_APP_TO_GQL_TOKEN,
+      token: secrets().APP.GITHUB_APP_TO_GQL_TOKEN,
```
gauntface: This could be secrets().GITHUB_APP.* dropping it from the name

PR #521, opened 2018-05-08
Fix post install

### LINKED ISSUE #519: Github App Post Install has a race condition
After installing the github app we redirect to post-install page that confirms the installation and redirects afterwards to the org config.

At the moment there is a race condition between this page and the webhook triggering and completing a setup step.

gauntface:
Fixes #519

### DIFF
--- a/src/server/dash-server.ts
+++ b/src/server/dash-server.ts
@@ -127,8 +127,8 @@ export class DashServer {
 
           const userRecord = await userModel.getUserRecordFromRequest(request);
           if (!userRecord) {
-            // User may not be signed but have project-health setup and
-            // have installed the app so we should redirect.
+            // User may not be signed but have project-health app installed,
+            // so we should redirect.
             response.redirect(
                 302,
                 `/signin?final-redirect=${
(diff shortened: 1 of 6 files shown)


src/server/dash-server.ts
```diff
@@ -123,13 +126,62 @@ export class DashServer {
             return;
           }
 
-          const installDetails = await githubAppModel.getInstallation(
-              Number(request.query.installation_id));
-          if (!installDetails) {
-            response.status(400).send('Invalid installation ID.');
+          const userRecord = await userModel.getUserRecordFromRequest(request);
+          if (!userRecord) {
+            // User may not be signed but have project-health setup and
```
samuelli: signed in but has the project-heath app installed, so we should redirect.

src/server/utils/generate-github-app-token.ts
```diff
@@ -21,19 +18,22 @@ export async function generateGithubAppToken(installId: number):
...
   const responseBody = await response.json();
   if (!responseBody.token) {
-    console.error('Unabel to g');
     throw new Error('Unable to generate to GitHub App auth token.');
   }
   return responseBody.token;
 }
 
-async function generateJWT(appId: number, certPath: string) {
+export async function generateJWT() {
```
samuelli: What is JWT? Can we rename/add comments?

PR #532, opened 2018-05-16
Fixes year string.

paullewis:
Currently the output says '1 years ago'. This should change that so it's '1 year ago', and 'n years ago' for all values of n that aren't 1 👍 

R @samuelli

### DIFF
--- a/src/client/public/scripts/dash/utils/time-to-string.ts
+++ b/src/client/public/scripts/dash/utils/time-to-string.ts
@@ -15,7 +15,7 @@ export function timeToString(dateTime: number) {
 
         if (secondsSince > 365) {
           secondsSince = secondsSince / 365;
-          unit = 'years';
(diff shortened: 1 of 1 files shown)


paullewis:
Fair! I broadened its horizons some.

paullewis:
@samuelli The build on Travis seems to fail because of OpenSSL, not something in my code. What's the best thing to do here?

samuelli:
> @samuelli The build on Travis seems to fail because of OpenSSL, not something in my code. What's the best thing to do here?

Fix it! If you can review https://github.com/PolymerLabs/project-health/pull/534 that should fix this PR.

samuelli:
@paullewis: oh I can't rebase it. Can you rebase this change? I guess I could disable the checks requirement and force merge it since it's completely fine.

PR #540, opened 2018-05-21
Migrate status to two separate notifications

samuelli:
This change builds on prior changes to migrate webhook handling to the subscriber model. The status check file has been moved and split into two separate notifications - one that handles auto merging and the other which notifies when failed status checks are encountered.


samuelli:
Good source diff: https://github.com/PolymerLabs/project-health/pull/540/files/231237f1b1eb6318e88d76eaacece451dc0373a6..fa070fbfbf899d167b72be42c2534b6613035807

Good test diff: https://github.com/PolymerLabs/project-health/pull/540/commits/fa070fbfbf899d167b72be42c2534b6613035807

PR #555, opened 2018-06-29
Remove race condition on app install

samuelli:
This change fixes production app credentials and also removes the race condition on app installation. As GitHub redirects immediately on app installation, regardless of whether or not the webhook event has been sent & handled, we were handling this installation in 2 places. As a result, in some cases, there was a race condition for writing to the database. This instead only handles the webhook and the redirect waits for an installation to be present.


gauntface (approved):
I suspect it might make sense at a later stage to move this logic to:

Use gets redirected to config after install, we realise it's not ready and send a "Still installing page, this shouldn't take long" page, this pings backend every second till it's ready and then refreshes.

With this, the timeout is pushed on to the server which means the server will be getting backed up while user is  looking at a spinner.

